PR #115, opened 2015-01-02
Modify RequestQuerystringEvent to pass request.RawUrl

asapostolov:
I use a different folder to store the images I need to be watermarked. So I need the RawUrl of the request to determine if a watermark is needed on that image.

Thanks for the great library!


JimBobSquarePants:
Cheers! I thought about then forgot about that when I was adding my implementation. :+1:

PR #118, opened 2015-01-13
Make the DLL COM visible

cosmo0:
Marks the DLL as COM-visible, so that we can use it in Ruby through `require 'win32ole'`.

I'm not sure why there are multiple commits on this merge, only the `AssemblyInfo.cs` file is modified.


JimBobSquarePants:
Could you reopen that pull requests against my v2 branch? It makes it easier to keep track of changes. Thanks :smile:

cosmo0:
hem creating a pull request on V2 results in even more changes in the PR, and updating it seems difficult... I'll have to clean up my copy.

cosmo0:
Well it turns out that we won't be using the DLL as COM after all, so...

JimBobSquarePants:
No worries chum.

PR #129, opened 2015-02-24
Removes unnecessary/unused files

cosmo0:
Just a little cleanup to remove the Web.Test project (that doesn't even compile) and a few unnecessary files (used by MSTest).


JimBobSquarePants:
Ace. Can you repoint that at v2? I was thinking of dropping all that anyway, could never get it to work properly.

cosmo0:
Darn. Since my branch was V2, I thought that it would make a PR on V2.

PR #131, opened 2015-02-25
Adds "inside rotation" filter/method

cosmo0:
Hey there

I need a method to rotate the images "inside their own area", so I just developed it.

For the parameter, I tried to keep consistency with the other methods asking for multiple values, so I created a `RotateInsideLayer` class.

Depending on the `KeepImageDimensions` parameter, it either keeps the original image dimensions, and zooms the rotated image data inside the image area (so as to not display any black area in the corners), or resizes the image area (to achieve the same effect).

The zoom Calculation has been taken from http://math.stackexchange.com/questions/1070853/

````csharp
using (ImageFactory imageFactory = new ImageFactory())
{
    imageFactory.RotateInside(new RotateInsideLayer {
        Angle = 45,
        KeepImageDimensions = false
    });
}
````

I also refactored the rotation a little bit, in order to extract the size calculation, so that it can be unit tested (and reused in this new method).

PS: sorry about the 3 commits that come with them (gitter badge, readme and version bump), I did a rebase on V2, so I don't understand why they're here...


cosmo0:
Awesome, I managed to remove the unwanted commits, and render the PR automergeable ! \o/

JimBobSquarePants:
You're better man with Git than I!

I'm wondering whether the methods within the `Rotation` class should be incorporated into [ImageMaths](https://github.com/JimBobSquarePants/ImageProcessor/blob/V2/src/ImageProcessor/Imaging/Helpers/ImageMaths.cs). That would make them more obviously reusable and testable. You'll find a `DegreesToRadians` method in there already.

I think `NewSizeAfterRotation` should probably be renamed to `GetBoundingRotatedRectangle` or something similar and return a `Rectangle` which the size can be determined from after. 

What do you think?

cosmo0:
I just forgot to `git fetch` before `git rebase`, so I fetched, `git rebase -i` to remove the unwanted commits, fix the merge conflicts, then `git push --force`.

I'll make the suggested modifications.

JimBobSquarePants:
`--force` Brave! I always feel like I'm gonna blow everything up when I use that.

Ace... I'm prepping for a release so if you have any more gems to add, let me know so I can get them in before I ship.

cosmo0:
Yes, the `--force` will make everything blow up if you're not alone on your branch. Since you hadn't answered yet and I worked alone, I felt it was OK. I also should have merged a few commits while I was at it, but hey.

I want to include the effects I wrote/copied from https://github.com/cosmo0/ImageProcessor.Plugins.Effects but I'm not sure I will be able to do it soon.

cosmo0:
There ! If you have any other remark, please do not hesitate :smile:

JimBobSquarePants:
Looks good. I'll get it merged.

Yeah, there's some great effects in your library. I could have a go moving them one at a time when I get a chance as there's a few bits in there that can be switched over to use FastBitmap.

JimBobSquarePants:
Just a heads up. I refactored a few thing in the library after the commit and as part of that I've renamed the method `RotateBounded`.

cosmo0:
There's a bug when the angle is negative :sweat: (black corners appear)

I'm on it.

PR #132, opened 2015-02-26
Fix rotatebounded

cosmo0:
Fixes the counter-clockwise bounded rotation (it zoomed out instead of zoomed in).

I modified the factory unit tests to save the output images so that I can see what happens, let me know if you want me to remove it.


JimBobSquarePants:
I think saving them for viewing is a great idea. I would like to make sure the path is added to gitignore though so we don't track them.

cosmo0:
Note that the calculations don't work when the angle is > 90° (or < -90°), which is normal given the basis of calculation, but it's not guarded.
I don't know if it's better to throw an error, leave the processor strange behavior (check out what it does with 140, 150 and 160 degrees), or fix it (it would be more logical to regular-rotate the image 90° then inner-rotate it, but hey, I'm not the user).

JimBobSquarePants:
The normal rotate class woks for any angle positive or negative angle so the behaviour should match. I'll have a look if I can get the chance.

Is it the zoom calculation that is broken or the rotate?

cosmo0:
Both, I think.
The zoom is based on this: http://math.stackexchange.com/a/1071959/11813 and at 180° it completely rotates the image (360°).

Since the zoom calculation depends on the `sin` and `cos` of the angle, when the angle is > 90, I think we should invert the `sin` and `cos` positions in the formula (I haven't tested yet, though).

But then, if we go this way, it should also work with angles after 180°, then after 270°, etc.

JimBobSquarePants:
Zoom isn't checking in both directions. I should be able to fix it.

JimBobSquarePants:
That should be it fixed. Can you double check for me?

cosmo0:
Yes, it works perfectly ! :+1: Thanks !

JimBobSquarePants:
I'm glad! Just released it :smile:

Issue #136, opened 2015-03-16
Question/Feature Request: Focal Point


Myster:
I would like to use resize mode=max and focal point (from Umbraco) but I cannot find elegant way to do this.

If the anchor parameter accepted x,y coordinates I think that would work perfectly.

JimBobSquarePants:
I'm not sure what you're after; those two won't work in combination. 

`ResizeMode.Max` resizes an image to the maximum possible width/height to fit the given dimensions whilst maintaining the original aspect ratio. Nothing is cut off or added so there is no anchoring involved.

Myster:
I figured it out, this does what I want.
`?mode=crop&center=0.2,0.3&width=100&height=10`
The http://imageprocessor.org/ documentation doesn't really cover this, I found it in the Umbraco source code. :-)

JimBobSquarePants:
That's `mode = crop` not `mode = max`. 

Docs are open source and I am but one man....

Myster:
Hi, I was going to add this to the docs but I can't find the `gh_pages` directory.

JimBobSquarePants:
Aw cheers! :+1: It's actually a branch on the repo not a directory.

Brace yourself with Jekyll though, it's a nightmare. The docs are a bit of a mess just now as I started converting everything to markdown but failed miserably.

https://github.com/JimBobSquarePants/ImageProcessor/tree/gh-pages

PR #138, opened 2015-03-20
Documented how to use center parameter

Myster:
Added documentation as per issue #136


JimBobSquarePants:
Thanks for that. Could you possibly fix the spelling of `piont`. :smile:

PR #143, opened 2015-03-25
If querystring isnt passed, don't throw exception.

perosb:
When trying to get remote images for example, this throws exception if no querystring is passed.


JimBobSquarePants:
Thanks for this. 
Could you possible repoint this to the developments branch v2.

perosb:
New pull request in #147

PR #154, opened 2015-04-21
Add NuGet dependency on ImageProcessor.Web.Plugins.AzureBlobCache

Jeavon:
If you install ImageProcessor.Web.Plugins.AzureBlobCache without ImageProcessor.Web.Config it doesn't work as it's not registered in web.config hence adding it as a dependency


JimBobSquarePants:
Doh!

PR #156, opened 2015-04-24
Remove unnecessary assignment/checks from Color32

Banane9:
Just something I noticed while looking into it..

The URI thing failing on `:` seems to make sense though btw, at least Windoes 8 doesn't allow usage of that char in paths.

### DIFF
--- a/src/ImageProcessor/Imaging/Colors/Color32.cs
+++ b/src/ImageProcessor/Imaging/Colors/Color32.cs
@@ -76,9 +76,6 @@ namespace ImageProcessor.Imaging.Colors
             this.R = red;
             this.G = green;
             this.B = blue;
-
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Yeah.. makes sense. 

URI thing?

Banane9:
Oh, the single failing test :)

JimBobSquarePants:
Ah yes, of course! :) That one had me thinking for a while, especially since it's configurable in your application also.

I''l merge this now.

Banane9:
Btw, if you want, I can help out a bit :)

JimBobSquarePants:
Absolutely! :) There's a plan to rewrite the library without system drawing so your input would be most valuable!

PR #157, opened 2015-04-25
create cache contain if it not exists, add work-a-round for dev storage emularto url

idseefeld:
add work-a-round for dev storage emularto url
create cache contain if it not exists

### DIFF
--- a/src/ImageProcessor.Web/Extensions/StringExtensions.cs
+++ b/src/ImageProcessor.Web/Extensions/StringExtensions.cs
@@ -138,17 +138,6 @@ namespace ImageProcessor.Web.Extensions
 
             return Uri.TryCreate(expression, UriKind.Relative, out uri) && uri.IsWellFormedOriginalString();
         }
-        /// <summary>
-        /// As url segments cannot contain : it will be replaced with _
-        /// However, we need the correct url for requesting local host on port 10000 where 
-        /// Development Storage Emulator is listening
-        /// </summary>
-        /// <param name="url">The <see cref="T:System.String">String</see> instance that this method extends.</param>
-        /// <returns>returns corrected url with host:port</returns>
-        public static string FixDevelopmentStorageEmulatorUrl(this string url)
-        {
-            return url.Replace("127.0.0.1_10000", "127.0.0.1:10000");
-        }
         #endregion
     }
 }
--- a/src/ImageProcessor.Web/Extensions/StringExtensions.cs
+++ b/src/ImageProcessor.Web/Extensions/StringExtensions.cs
@@ -138,6 +138,17 @@ namespace ImageProcessor.Web.Extensions
 
             return Uri.TryCreate(expression, UriKind.Relative, out uri) && uri.IsWellFormedOriginalString();
         }
+        /// <summary>
+        /// As url segments cannot contain : it will be replaced with _
+        /// However, we need the correct url for requesting local host on port 10000 where 
+        /// Development Storage Emulator is listening
+        /// </summary>
+        /// <param name="url">The <see cref="T:System.String">String</see> instance that this method extends.</param>
+        /// <returns>returns corrected url with host:port</returns>
+        public static string FixDevelopmentStorageEmulatorUrl(this string url)
+        {
+            return url.Replace("127.0.0.1_10000", "127.0.0.1:10000");
+        }
         #endregion
     }
 }
--- a/src/ImageProcessor.Web/Services/RemoteImageService.cs
+++ b/src/ImageProcessor.Web/Services/RemoteImageService.cs
@@ -18,6 +18,7 @@ namespace ImageProcessor.Web.Services
     using System.Web;
 
     using ImageProcessor.Web.Helpers;
+    using ImageProcessor.Web.Extensions;
 
     /// <summary>
     /// The remote image service.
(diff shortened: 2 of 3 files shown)


src/ImageProcessor.Web/Extensions/StringExtensions.cs
```diff
@@ -138,6 +138,17 @@ public static bool IsValidVirtualPathName(this string expression)
...
             return Uri.TryCreate(expression, UriKind.Relative, out uri) && uri.IsWellFormedOriginalString();
         }
+        /// <summary>
+        /// As url segments cannot contain : it will be replaced with _
+        /// However, we need the correct url for requesting local host on port 10000 where 
+        /// Development Storage Emulator is listening
+        /// </summary>
+        /// <param name="url">The <see cref="T:System.String">String</see> instance that this method extends.</param>
+        /// <returns>returns corrected url with host:port</returns>
+        public static string FixDevelopmentStorageEmulatorUrl(this string url)
```
JimBobSquarePants: I'm not completely happy adding something like this. By all means we should be able to use a url with a port but that should be something that is url encoded by the dev and then decoded in the HttpModule for IImageService implementations to use.  

I've also used this url/port combination before for hosting projects I have in development in IIS so I don't want potential conflict.

src/ImageProcessor.Web/Services/RemoteImageService.cs
```diff
@@ -134,7 +135,9 @@ public bool IsValidRequest(string path)
         /// </returns>
         public async Task<byte[]> GetImage(object id)
         {
-            Uri uri = new Uri(id.ToString());
+            Uri uri = new Uri(id.ToString()
```
JimBobSquarePants: Again we should url encode/decode instead.

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -287,7 +302,9 @@ public AzureBlobCache(string requestPath, string fullPath, string querystring)
                 {
                     string container = RemoteRegex.Replace(this.cloudSourceBlobContainer.Uri.ToString(), string.Empty);
                     string blobPath = RemoteRegex.Replace(this.RequestPath, string.Empty);
-                    blobPath = blobPath.Replace(container, string.Empty).TrimStart('/');
+                    blobPath = blobPath.Replace(container, string.Empty).TrimStart('/')
+                        //.Replace("127.0.0.1_10000", "127.0.0.1:10000")
+                        .FixDevelopmentStorageEmulatorUrl();
```
JimBobSquarePants: As above.

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -66,6 +66,21 @@ public class AzureBlobCache : ImageCacheBase
         private string cachedRewritePath;
 
         /// <summary>
+        /// gets reference or creates container
+        /// </summary>
+        /// <param name="cloudBlobClient">cloud blob client</param>
+        /// <param name="containerName">name of container</param>
+        /// <param name="accessType">access permissions ie. public</param>
+        /// <returns>container for cache items</returns>
+        private CloudBlobContainer CreateContainer(CloudBlobClient cloudBlobClient, string containerName, BlobContainerPublicAccessType accessType)
```
JimBobSquarePants: This is a great addition! I didn't know you could do that.

However, putting my StyleCop hat on :cop:, private methods should always be below public ones and I specifically avoid using var for anything other than anonymous types in this project.

JimBobSquarePants:
Hi Dirk, Cheers for this, It's great to have you contributing :smile: 

Don't be put off by the comments, I'm just super picky with this project. I think with a few tweaks we can have a great enhancement to the library.

src/ImageProcessor.Web/Services/RemoteImageService.cs
idseefeld: You are right. That would be much better. I try that out within the next days...

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
idseefeld: Again, I agree - should have read your recommendations for contributions more carefully. However, I hope you got the idea ;-)

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
idseefeld: I definatly have to read more about this async stuff. Currently I am not very familiar with it. But your project is a good starting point and inspiration! Thank you very much! :+1:

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
JimBobSquarePants: Absolutely, I think it's ace.

idseefeld:
Hi James, I have reworked my pull request. I hope your are satisfied with the update. Of course I DO understand your concerns. And I am 100% on your side being super picky. It is quite obvious how much effort you put into this project. I thank you very much for your detailed comments!

JimBobSquarePants:
Thanks my friend. I'll have a look as soon as possible.

JimBobSquarePants:
My apologies but I had to revert the merge as it was made against the wrong branch. 

Could you resubmit it against the V2 branch? That is the current default.

PR #163, opened 2015-05-06
Used the Unvalidated version of RawUrl

evdinursan:
Accordingly to this SO issue (http://stackoverflow.com/questions/30075396/http-module-intercept-requests-and-breaks-custom-errors-configuration), if I was trying to use a special URL to simulate an attack on the webapp (for example, .../App/<c), the custom error pages didn't reacted anymore.


JimBobSquarePants:
Cheers for this. Much appreciated!

PR #177, opened 2015-06-11
Remove "preset=" from queryString

stefbp:
There was a problem with the preset parameters. After changing the preset name in the Query String, the "preset =" value persisted. This problem meant that the preset settings were not all supported by ImageProcessor.

### DIFF
--- a/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
+++ b/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
@@ -555,7 +555,7 @@ namespace ImageProcessor.Web.HttpModules
 
                         // We use the processor config system to store the preset values.
                         string replacements = ImageProcessorConfiguration.Instance.GetPresetSettings(preset);
-                        queryString = Regex.Replace(queryString, preset, replacements ?? string.Empty);
+                        queryString = Regex.Replace(queryString, match.Value, replacements ?? string.Empty);
                     }
                 }
             }


JimBobSquarePants:
I'm sorry but can you explain that again. What are you doing?

stefbp:
Yes, it's simple. There is a problem with the presets feature.

In the config file "processing.config", if I add this preset:

    <preset name="square" value="width=200&#038;height=200&#038;mode=crop&#038;upscale=false" />

and then I call the `` it doesn't work. The only firsts parameters of the preset is processed.

So, I debug your code and I found this line into method `ReplacePresetsInQueryString`:

    queryString = Regex.Replace(queryString, preset, replacements ?? string.Empty);

The result was:

**input queryString:** "[...]?mode=pad&rnd=[...]&preset=square"

**ouput queryString:** "[...]?mode=pad&rnd=[...]&**preset=**width=200&height=200&mode=crop&upscale=false"

Please remark the parameter "preset=width=200[...]"

It's the reason why I change your code to


    queryString = Regex.Replace(queryString, match.Value, replacements ?? string.Empty);

Now, this is the result:

**input queryString:** "[...]?mode=pad&rnd=[...]&preset=square"

**ouput queryString:** "[...]?mode=pad&rnd=[...]&width=200&height=200&mode=crop&upscale=false"

there is no more "preset=" and the preset is working like a charm!

JimBobSquarePants:
Thanks for this. I must have broken it when I rewrote the parser.

Issue #161, opened 2015-05-01
AzureBlobCache plugin use redirect instead to rewrite


abadyl:
Hi, just installed AzureBlobCache plugin. It work well, but with one remark.

When I access images, I all time redirected to CDN url. That mean if my image already in cache, server will redirect me to Azure CDN url. 

This is extra round trip, and also issue with SEO for images.

Is it possible to not use redirect, but make call for image to azure blob via API, get image stream, and then copy it to output stream of response?

In this case we will have clean SEO for images, and without extra redirect.

What was an idea to use redirect?

Thanks,

JimBobSquarePants:
Hi there,

Could you clarify in what way this would be a SEO issue?

If I were to request the urls directly from the CDN Google would treat that in the same way as my 302 redirect from the request url. (I use 302 because if the source file is updated with the same name, the request url could be the same but the CDN url would be different.)

There's no extra round trip; the url is intercepted and redirected. I have to use a url hosted within the currently running application for all image requests because there is no way to intercept an outgoing request to another domain.

abadyl:
Hi, 
I agree that 302 redirect will work from SEO point.

But, if you have on a page i.e. 100 images => browser will be redirected 100 times. I also understand that if we will user real pull CDN it is not an a problem, but if we not use CDN at all. Or if your container should be private?

It also difficult to tell  how it (redirects) will influence on overall performance of website.

Why you not use API or webrequest to get cache stream from server and write it to response?

Will code bellow make sence?

        
        public void RewritePath(HttpContext context)
        {
            HttpWebRequest httpWebRequest = (HttpWebRequest)WebRequest.Create(this.cachedRewritePath);
            httpWebRequest.Method = "HEAD";

            using (HttpWebResponse httpWebResponse = (HttpWebResponse)httpWebRequest.GetResponse())
            {
                //get image stream from cloud
                Stream ourStram = httpWebResponse.GetResponseStream();

                ourStram.CopyTo(context.Response.OutputStream);
            }
        }

JimBobSquarePants:
Interesting... Would you have the opportunity to fork and test this? 

I'm incredibly busy just now so you would most probably be faster than me.

abadyl:
Sure, this weekend I will try to play with it.

Will keep you informed.

JimBobSquarePants:
Brilliant! Thanks for your input :smile:

abadyl:
Hi, 
[...]

PR #182, opened 2015-06-17
Allow streaming or redirecting of cached image via Azure

gnpretorius:
Added the functionality mentioned here

https://github.com/JimBobSquarePants/ImageProcessor/issues/161

I tested it locally and all looked good but it would be worth having a quick look and testing your side. 

If no config setting is present, the behaviour stays as is. If the setting is present and is set to anything other than true/True, the behaviour stays. Otherwise, it streams the image.

<setting key="StreamCachedImage" value="true" />

Any issues, let me know

G


JimBobSquarePants:
Thanks for this. I'll have a good loo at it and test asap.

PR #183, opened 2015-06-17
Add overload for loading image data from a byte array.

khiemtong:
I found myself writing a lot of code similar to this

```
var bytes = File.ReadAllBytes(...)
using (var memStream = new MemoryStream(bytes)) { // extraneous
   ...
   imageFactory.Load(memStream)
}
```

This overload removes the need to create an extra memory stream, since ImageFactory.Load already contains ones internally.

### DIFF
--- a/src/ImageProcessor.UnitTests/ImageFactoryUnitTests.cs
+++ b/src/ImageProcessor.UnitTests/ImageFactoryUnitTests.cs
@@ -77,6 +77,23 @@ namespace ImageProcessor.UnitTests
             }
         }
 
+		[Test]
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Good thinking! Thanks!

Just a heads up, I think you are using different formatting in Visual Studio than I. I'd keep an eye on that when submitting pull requests.

Cheers!

JimBobSquarePants:
Added in version [2.2.6](https://www.nuget.org/packages/ImageProcessor/) Thanks for your contribution is is very much appreciated :+1:

Issue #192, opened 2015-07-07
NullReferenceException in ImageProcessor.Web.Helpers.ExtendedColorTypeConverter


AndyA13:
I can't reproduce this locally :rage:  but keeps happening on an azure website instance.

Not 100% sure the issue is at your end, but you probably know more about the type converters and can shine some light on it.

`ImageProcessor.Web.Helpers.ExtendedColorTypeConverter in ConvertFrom at line 66:0`

The line number doesn't seem to be accurate.

When creating a spreadsheet using [ClosedXml][0], the [`InitializeTheme` method][1] makes some calls to [`XLColor.FromHtml`][2]

It is using 8 character hex codes :anguished: Examples:
```
    Text1 = XLColor.FromHtml("#FF000000"),
    Background1 = XLColor.FromHtml("#FFFFFFFF"),
    Text2 = XLColor.FromHtml("#FF1F497D"),
    Background2 = XLColor.FromHtml("#FFEEECE1")
```
It looks like `XLColor.FromHtml(string htmlColor)` is calling `ColorTranslator.FromHtml(htmlColor)` which is finding your type converter and calling that.

Stacktrace:
```
    Module "Usercontrols.FormDataViewerPages.List", line 6, in DataTableToExcelByteArray
        Byte[] DataTableToExcelByteArray(System.Data.DataTable, System.String)
    Module "ClosedXML.Excel.XLWorkbook", line 245, in .ctor
        Void .ctor(ClosedXML.Excel.XLEventTracking)
    Module "ClosedXML.Excel.XLWorkbook", line 0, in InitializeTheme
        Void InitializeTheme()
    Module "System.Drawing.ColorTranslator", line 297, in FromHtml
        System.Drawing.Color FromHtml(System.String)
    Module "ImageProcessor.Web.Helpers.ExtendedColorTypeConverter", line 66, in ConvertFrom
        System.Object ConvertFrom(System.ComponentModel.ITypeDescriptorContext, System.Globalization.CultureInfo, System.Object)
```

A quick local test with `ColorConverter` seems to return a valid color from the 8 character codes.

Packages | Versions
-------------|-------------
DocumentFormat.OpenXml | 2.5.5631.0
ClosedXml | 0.76.0.0
ImageProcessor | 2.2.6.0
ImageProcessor.Web | 4.3.4.0

[0]: https://closedxml.codeplex.com/
[...]

PR #194, opened 2015-07-08
Fix/8 digit hex color codes

AndyA13:
For issue #192 

This fixes the issue, but I guess there may be more work to do with moving null culture checks down into the other type converters.


JimBobSquarePants:
That looks great to me thanks! I might make a base converter or simply add the null check for each converter to handle the other issue..

PR #202, opened 2015-08-12
Fix 404 error handling in RemoteFile

mbengtson:
The response is always null when a 404 WebException is thrown, so the error handling was never performed.
Using ex.Response works as expected.


mbengtson:
BTW... do you have any suggestions for how to serve a default image as fallback when the remote service returns a 404 error?

JimBobSquarePants:
Good spot thanks! I really appreciate your help. :smile: 

The best way would be to write your own `IImageService` implementation to replace the remote file service. On 404 you could load up your replacement image and send the bytes from that to the processor.

You'll need to install the ImageProcessor.Web.Config Nuget package to replace the remote service. The configuration details would then be found in `security.config`

Issue #211, opened 2015-09-08
the term "media" is globally replaced in path


perosb:
It seem thata the "prefix", ie "media" is globally replace in the path which means that you cannot have an image named ```my_media_image.png```

perosb:
my git is strange and not sure what up.

fix here https://github.com/perosb/ImageProcessor/commit/b1d1e6523f8593448ac56dfc7824b9cc96e77ed8

JimBobSquarePants:
Fixed in [v4.4.0](https://www.nuget.org/packages/ImageProcessor.Web/4.4.0)

PR #212, opened 2015-09-08
Fix for media replacement

perosb:
try to fix https://github.com/JimBobSquarePants/ImageProcessor/issues/211


perosb:
I really suck at git'ing.

JimBobSquarePants:
It got me a bit at first. The trick is to fetch from the correct origin branch before updating and submitting a pull requests. This will automate most of that for you. https://desktop.github.com/

PR #213, opened 2015-09-13
Added MetaTests.cs, updated csproj.

christopherbauer:
Moq doesn't seem to be installed and ImageFactory is not derived from an interface so I find writing a meaningful unit test for ProcessImage to be difficult. Any thoughts on extracting an interface for testing purposes? If you agree, I can add in a meaningful unit test for that with a new pull request (unless you want integration tests).

### DIFF
--- a/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
+++ b/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
@@ -68,6 +68,7 @@
     <Compile Include="Imaging\FastBitmapUnitTests.cs" />
     <Compile Include="Imaging\Filters\Photo\MatrixFilterBaseTests.cs" />
     <Compile Include="Imaging\Helpers\ImageMathsUnitTests.cs" />
+    <Compile Include="Processors\MetaTests.cs" />
   </ItemGroup>
   <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Thanks for this. Very much appreciated!

Yeah, that's been a problem for me throughout all this. I should really have defined a simple interface for ImageFactory (Load,Save, etc) which would make testing a lot easier. When I first started development I was fairly inexperienced and never thought this would take off so there's some architectural issues.

I don't think adding an interface now would cause any issues and am happy to do so to improve testability but a second opinion would be great. 

If you don't think it's wise I've no issues with adding Moq if need be.

christopherbauer:
Well, if you want unit tests for these methods we'd have to! But it's not strictly necessary to have unit tests at all, and the methods are simple. We could just write integration tests anyway. Up to you really :+1:

JimBobSquarePants:
I'm gonna go ahead and merge this then look at an interface.

JimBobSquarePants:
I just had a go at extracting an interface.... It was going swimmingly until I remembered that I have two properties with private and internal setters. `Image` and `CurrentGamma`. The latter I don't need now but I'm not all that sure I want to make the `Image` property setter public.

What are your thoughts?

christopherbauer:
You can do (sorry if this isn't exactly right)
    
    public interface IImageFactory
    {
        System.Image Image { get; }
    }

And Moq has a method called SetupGet() that will allow us to create unit tests against it without having to expose the setter.

JimBobSquarePants:
If I was to extract an `IImageFactory` interface within ImageProcessor I would have to change the signature of any method or interface referencing `ImageFactory` would have to change to reference the interface instead. Having only a getter won't allow me to set the property.

We also have the issue that changing all the signatures to point to the interface *could* potentially break any extension methods anyone has written and would certainly break any `IGraphicsProcessor ` implementations. 

I can't really justify that now without a major version number which this release shouldn't have.

christopherbauer:
Good point - I definitely would not want to break anyone else's code. This can be put off to a later date and functional tests can be written to complete the coverage requirements.

JimBobSquarePants:
Cool. v3 WIll be interface driven anyway. I've learned my lesson :wink:

PR #216, opened 2015-09-14
Fix for ImageMaths.RotatePoint + added tests

christopherbauer:
Original code was using a bad stackoverflow answer that didn't work when you tried to rotate around a center point that wasn't 0, 0. This can be observed by running the new unit tests in the old version of the code.

### DIFF
--- a/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
+++ b/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
@@ -78,34 +78,34 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="packages.config" />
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
And this is exactly why I wanted help on the unit tests! Great spot and well done! 

I'll get all these merged later on once I've had a good read.

christopherbauer:
Hah Jim, I just found out the reason for this - the parenthesis were wrong in the original implementation. I found this out by warning people off the stackoverflow answer, the answerer being rude, and me retesting the impl using a direct copy-paste from his answer. If you like, I can just fix the parenthesis and we can continue using the old code.

JimBobSquarePants:
Bugger I must have done that then. If you could that would be awesome thanks. :+1:

PR #218, opened 2015-09-15
Added bounding rectangle method tests

christopherbauer:
Names for these tests might be a little wonky.


JimBobSquarePants:
We can always rename them in the future. :smile:

PR #219, opened 2015-09-17
Added to points unit tests

christopherbauer:
Simple tests


JimBobSquarePants:
Thanks for this... Think there might be some conflicts though. Could you double check your local version is up to date?

christopherbauer:
Yeah I was in the midst of it, still trying to get the whole git workflow thing to make sense. Long day :)

JimBobSquarePants:
Must be very late where you are no? Try to force yourself to always fetch before updating.

PR #223, opened 2015-09-18
Added some directory info extension unit tests.

christopherbauer:
Please verify that the setup methods do not cause horror and/or dismay

### DIFF
--- a/src/ImageProcessor.Web.UnitTests/Extensions/DirectoryInfoExtensionsUnitTests.cs
+++ b/src/ImageProcessor.Web.UnitTests/Extensions/DirectoryInfoExtensionsUnitTests.cs
@@ -1,0 +1,176 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="DirectoryInfoExtensionsUnitTests.cs" company="James South">
+//   Copyright (c) James South.
+//   Licensed under the Apache License, Version 2.0.
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace ImageProcessor.Web.UnitTests.Extensions
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+
+    using Web.Extensions;
+
+    using NUnit.Framework;
+
+    /// <summary>
+    /// The directory info extensions unit tests.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Cheers. Will do! :smile:

christopherbauer:
Wasn't sure if this would be here in the morning or merged, but I figure I'd explain just a bit;

Didn't pull the setup logic out to a separate function because I'm pretty strict about "if you need it 3x refactor, else keep it separate" mentality. 
These operations are setup/teardowns so the tests are repeatable, and if directory info is extended for other things such as moving a directory I don't want those tests to impact one-another.

src/ImageProcessor.Web.UnitTests/Extensions/DirectoryInfoExtensionsUnitTests.cs
```diff
@@ -0,0 +1,176 @@
...
+            /// </summary>
+            private IEnumerable<string> directoryList;
+
+            /// <summary>
+            /// The setup directories.
+            /// </summary>
+            [SetUp]
+            public void SetupDirectories()
+            {
+                this.directoryList = Enumerable.Range(1, this.directoryCount).Select(i => $"{TestDirectoryRoot}/TestDirectory{i}");
```
JimBobSquarePants: Didn't know you could do this. Looks like the new C#6 string templating.

christopherbauer: Woops it is, I didn't even think about it. We probably shouldn't be doing that yet huh?

JimBobSquarePants: :smile: ....I thought about switching over but decided it was too risky for anyone collaborating. I'll need to check my AppVeyor settings as that didn't pick up on the issue.

christopherbauer:
I rolled back the string formatting to be compliant to .net 4. String interpolation was fun while it lasted! Do you prefer squash or just rolling the whole thing in as-is?

JimBobSquarePants:
I'm gonna merge the whole thing as is. We can always refactor tests after.

PR #226, opened 2015-09-22
Added lanczos kernel test

christopherbauer:
I left a few cases commented out - these are the answers the current implementation returns, but I can't independently verify their correctness because Lanczos is beyond my grasp. If you have a method of independently verifying them, or someone smarter than I can double check the expected values are correct, please proceed.


JimBobSquarePants:
I'll have a good look at it thanks!

JimBobSquarePants:
I'm deleting all the custom resize code including Lanczos  as it was implemented incorrectly. I have working version and tests in v3

PR #228, opened 2015-09-22
Added color moment tests

christopherbauer:
Weird thing I noticed with these; weighted distance comes out as a long and chops off the decimal (effectively rounding down) however the variance uses the decimal number and returns a float. If all that sounds okay then we're good to go!


JimBobSquarePants:
Good find.... This is something I should investigate. I inherited the code for this quantizer from another project. There should definitely be greater consistency across returned values.

PR #231, opened 2015-10-05
Added CmykColorTests.cs

christopherbauer:
More tests need to be added.

Due to the the static analyzer saying some of the implicit operators were unusable due to ambiguousness, and per http://stackoverflow.com/questions/25775645/ambiguous-implicit-user-defined-conversions-in-net, I removed some implicit operators that were converting to other colors types.

### DIFF
--- a/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
+++ b/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
@@ -65,6 +65,7 @@
     </Compile>
     <Compile Include="Extensions\DoubleExtensionsUnitTests.cs" />
     <Compile Include="Extensions\IntegerExtensionsUnitTests.cs" />
+    <Compile Include="Imaging\Colors\CmykColorTests.cs" />
     <Compile Include="Imaging\ColorUnitTests.cs" />
     <Compile Include="Imaging\CropLayerUnitTests.cs" />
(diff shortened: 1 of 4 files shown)


christopherbauer:
Hey Jim - take a look at the implicit operators I removed here - the static analyzer basically threw up whenever I tried to invoke them during testing. I did a little research and turned up the stack overflow post that explains why and what probably should be done about it and went from there. What do you think?

JimBobSquarePants:
I'll have to have a proper look at this here as I have used implicit operators for V3. 

Could you have a look at the color code in there to see if it suffers the same issue?

christopherbauer:
In v3 it looks like HSV and YCbCr both have the conversion to BGRA in their source. According to the stack overflow post/my understanding of SRP, this is kind of breaking encapsulation.

All we would have to do is move those functions to BGRA and not implement the implicit conversions to X in Y class. Keeping the responsibility to implicitly convert to itself alongside the class that you're converting to seems to follow SRP to me, but it's your lib :)

JimBobSquarePants:
Ok... That kinda makes sense to me also. The bi-directional implicit casting I have right now would certainly be confusing to users as it stands and may lead to error. I really want to get this right.

christopherbauer:
Yeah, in that case this one can be merged in and I can look at adding tests for the remaining implicit operators and cleaning them up as needed.

JimBobSquarePants:
There we go.. :)

Issue #236, opened 2015-10-08
Resizing with Box Padding with width and height when only one dimension is less that target dimensions results in cropped image


x2764tech:
I may have read the documentation incorrectly, so please correctly if I'm wrong...

I have an image that is `94x107`: 
![94x107](https://cloud.githubusercontent.com/assets/2870080/10364163/1236e60c-6db1-11e5-83ba-aa3aa36d7713.png)

When I resize it to `100x100` with `ResizeMode.Pad` I get (the magenta is there so you can see what is going on)
![94x107-pad](https://cloud.githubusercontent.com/assets/2870080/10364222/6dad5336-6db1-11e5-9d5e-3a2fc24f8ef4.png)

However, when I resize it to `100x100` with `ResizeMode.BoxPad`, I get:
![94x107-boxpad](https://cloud.githubusercontent.com/assets/2870080/10364233/879456be-6db1-11e5-97a9-2b7de346bcff.png)

I believe this is due to the test on [Line 237 of Resizer.cs](https://github.com/JimBobSquarePants/ImageProcessor/blob/V2/src/ImageProcessor/Imaging/Resizer.cs#L237) which tests if *either* dimension would result in upscaling, rather than if *both* dimensions would result in upscaling (if I flip the `||` to an `&&`, I get the correct results).

I can submit a pull request if it would help, but I just wanted to run this by you before I did so.

JimBobSquarePants:
Thanks! Looks like you're on to something here. The behaviour when downscaling a dimension should match the behaviour of `ResizeMode.Pad`. I'll have a look at the PR as soon as possible.

christopherbauer:
Hey there @x2764tech , I see you wrote a 'stubbed resizer' in your test implementation and tested that class. If you want to have this PR potentially merged, it would be better to write a test for what you are expecting the output to be, changing the implementation to fix the error, then committing that. At that point I can take a look at it and merge it in.

JimBobSquarePants:
Thanks for the PR, tests are passing and fix is in place. Much appreciated! :+1:

JimBobSquarePants:
Fixed in the new release available on Nuget.

x2764tech:
Thanks!

PR #237, opened 2015-10-08
Feature/unit test for resizing

x2764tech:
Adding some tests to exercise issue #236


christopherbauer:
Hey there @x2764tech , I see you wrote a 'stubbed resizer' in your test implementation and tested that class. If you want to have that PR potentially merged, it would be better to write a test for what you are expecting the output to be, changing the implementation to fix the error, then committing that. At that point I can take a look at it and merge it in.

x2764tech:
Hey @christopherbauer - 
Yeah - I can send a fix for the broken test too - I put these tests in so that "you" could verify #236 - I wasn't sure that I was expecting the correct result.

BTW - the stubbed class is there to store the parameters to the `protected` method call...

Issue #240, opened 2015-10-15
ReplaceColor doesn't use the chosen color


dentalsoftware:
ReplaceColor doesn't use the chosen color. E.g. instead of drawing things yellow as set, the color shows purple. This behavior has changed in the latest release 2.3.0.

regards
Fabian

christopherbauer:
Can you post the code you're using so I can take a look?

dentalsoftware:
Hello Christoph

Here is what I do:

        With myProcessor
            Try
                .Load(myInStream)
            Catch ex As Exception
                MsgBox(ex.Message)
                Return False
            End Try
                .ReplaceColor(Color.LightGray, Color.Yellow, 10)
            End Select
            .Save(myOutStream)
            myOutStream.Flush()
	End With
I attach 2 sample images before and after Color changement...

Thanks and regards
Fabian
![before](https://cloud.githubusercontent.com/assets/15136812/10513271/c0677fa8-7346-11e5-8439-647861f1067c.png)![after](https://cloud.githubusercontent.com/assets/15136812/10513270/c065223a-7346-11e5-89b9-5f5455b41318.png)

christopherbauer:
OK - 2.3.0 was released a while ago (august 2013). Did you update to that version or did you update the most recent version of the main repo?

dentalsoftware:
Hello Christopher,

 

This is the latest release:

 



 

Kind regards

Fabian 

 

 

Von: Christopher Bauer [mailto:notifications@github.com] 
Gesendet: Donnerstag, 15. Oktober 2015 19:32
An: JimBobSquarePants/ImageProcessor
Cc: dentalsoftware
Betreff: Re: [ImageProcessor] ReplaceColor doesn't use the chosen color (#240)

 

OK - 2.3.0 was released a while ago (august 2013). Did you update to that version or did you update the most recent version of the main repo?

christopherbauer:
Sorry Fabian, there doesn't seem to be any content after your ':'. What version were you on and what version did you update to?

dentalsoftware:
I posted a screen shot from the ImageProcessor site (https://www.nuget.org/packages/ImageProcessor/). Here the same in textual form:

 

ImageProcessor               2.3.0      Sunday, September 20 2015

 

In the previous version (2.2.8) replacing a color worked perfect.

 

Kind regards

Fabian

 

Von: Christopher Bauer [mailto:notifications@github.com] 
[...]

PR #246, opened 2015-10-23
Replace color issue fix for #240

christopherbauer:
I think this code may have been copy pasted in from elsewhere which might explain why instead of performing a replacement it appears to be trying to determine the difference between the origin pixels.


src/ImageProcessor.Playground/Program.cs
```diff
@@ -35,48 +33,12 @@ public class Program
...
-
-                // Report back.
-                long peakWorkingSet64 = Process.GetCurrentProcess().PeakWorkingSet64;
-                float mB = peakWorkingSet64 / (float)1024 / 1024;
-
-                Console.WriteLine(@"Completed {0} in {1:s\.fff} secs {2}Peak memory usage was {3} bytes or {4} Mb.", fileInfo.Name, stopwatch.Elapsed, Environment.NewLine, peakWorkingSet64.ToString("#,#"), mB);
-            }
-
-            Console.ReadLine();
+            var files = GetFilesByExtensions(new DirectoryInfo(Environment.CurrentDirectory).GetDirectories("Images").Single(), ".png");
```
JimBobSquarePants: Can we make sure the playground code is not altered on merge? The timing and memory diagnostics code is very useful here.

JimBobSquarePants:
Can you post the transformed test image running this code?

christopherbauer:
test

src/ImageProcessor.Playground/Program.cs
christopherbauer: Yeah I'll roll that back in the branch when the fix is considered complete.

christopherbauer:
Hmm looking at the two images, that snippet I killed was doing more than I thought. I'll have to take a second look.

PR #248, opened 2015-11-15
Feature/v3 build automation

Jeavon:
Changed both projects to xproj
Followed Xunit guidelines to execute the tests using dnx v5
Add appveyor.yml which will build the NuGet package and execute the Xunit tests
Added a small NodeJs script (in build folder) executed by AppVeyor to update the version info in project.json to AppVeyor build, currently fixed to using "-alpha" suffix

I doubt this is the perfect setup but based on the information I could find it seems to be the right way to do it and it will allow you to create a MyGet feed of current builds.

Requires DNX beta8


JimBobSquarePants:
Bugger! I got all excited and was about to merge this then I realised that xproj is the ASP specific project type. That's definitely wrong for my needs. ImageProcessor is designed for use within all the new platforms.

If you look at the example in corefx here https://github.com/dotnet/corefx/tree/master/src/System.Collections.Concurrent/src

MS are using csproj which I had. They're even using MS build version 12.

Jeavon:
I think that xproj became generic in beta6 (but I'm not sure). I tired to follow this DNX project guide as much as possible http://docs.asp.net/en/latest/dnx/projects.html

JimBobSquarePants:
I'll double check in the gitter chat for corefx today to be sure. Thanks for all this anyway. Really looking forward to getting it all together.

Jeavon:
Good plan, it's as clear as mud. I'm happy I can make it work with either now I understand DNX a lot better

JimBobSquarePants:
I think I'm following what you did a bit better now. You used the "Class Library Package" project type yeah? That used to be ASP only but now isn't. (Explorer still thinks it is confusingly). That makes much more sense now. The only thing confusing me is `aspnet5-version.js` I take it that is a misnomer?

Jeavon:
Yes and yes. I will rename that file, it was just what the AppVeyor guys suggested it was called. I have realised that DNX doesn't actually use the xproj or csproj it's all about the project.json. 
One of the challenges is that a lot has changed through rhe betas, it wasn't even called DNX that long ago!

Jeavon:
The reason for using the xproj is that there is checkbox for build output which creates the NuGet package.

JimBobSquarePants:
I will admit I'm a lot more happier digging around in a .json file than a .csproj file. I'm sure in a year or so I reckon we'll be looking back thinking how we ever coped with the old setup.

Jeavon:
I'm sure you are right!

Jeavon:
Renamed aspnet5-version.js to appveyor-project-version-patch.js

Jeavon:
Also I'm 90% sure that project.json.lock should be excluded from source control e.g. https://github.com/aspnet/dnx/blob/dev/.gitignore

JimBobSquarePants:
I think you're right. It's been nothing but a burden.

PR #250, opened 2015-11-17
Adding a TryFiveTimes method for AzureBlobCache, currently only using…

Jeavon:
… within the RewritePath method but could/should be used everywhere that a connection is made


JimBobSquarePants:
Couldn't this lead to really slow page loads? Those image requests can sometimes be blocking depending on how many are on the page.

JimBobSquarePants:
Cool... I'll experiment today and see how it gets on with my current crappy cdn behaviour.

JimBobSquarePants:
Merged. Thanks for that :+1:

PR #252, opened 2015-11-24
Specify that MSBuild v14 use CoreClr RC1 for build

Jeavon:
The forces AppVeyor MSBuild to use CoreClr x64 RC1 when building ImageProcessor.sln

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -27,10 +27,9 @@ before_build:
 
     cd..
     
-build: 
-  project: ImageProcessor.sln
-  verbosity: minimal
-  
+build_script:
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Moar stuff? How did you figure this one out?

Jeavon:
Although it's not strictly necessary, on OSX it was building with CoreClr x64 so I wanted to ensure it was doing the same on AppVeyor on Windows. Did some digging on the MSBuild GitHub repo https://github.com/microsoft/msbuild They are currently porting all of MSBuild to CoreClr, wow!

JimBobSquarePants:
Amazing!

PR #254, opened 2015-11-26
fix #253. ImageFactory.Load(Stream) Does not support unseekable streams.

### LINKED ISSUE #253: ImageFactory.Load(Stream) Does not support unseekable streams.
In case of `stream.CanSeek == false`. `ImageFactory.Load(Stream)` throws argument exception.

Image.FromStream(memoryStream) throws  'System.ArgumentException' "Parameter is not valid"

Looks like problem is in `FormatUtilities.GetFormat(stream)` because it affects input stream instead of seekable memoryStream.

musukvl:
`FormatUtilities.GetFormat` works on memoryStream because it seekable.

`memoryStream.Position = 0;` removed because `FormatUtilities.GetFormat` does it.

### DIFF
--- a/src/ImageProcessor/ImageFactory.cs
+++ b/src/ImageProcessor/ImageFactory.cs
@@ -174,13 +174,6 @@ namespace ImageProcessor
         /// </returns>
         public ImageFactory Load(Stream stream)
         {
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Ah... good catch and solution. Thanks!

Issue #268, opened 2015-12-04
jpegtran errors with: "This program is blocked by group policy."


nul800sebastiaan:
This certainly is some interesting code.. ;)
https://github.com/JimBobSquarePants/ImageProcessor/blob/c4a50cb6fc47b4ca5e725b0e5c1ca76855efaf2a/src/ImageProcessor.Web/PostProcessor/PostProcessor.cs#L150

It's breaking all images for some people who have a limiting security policy set by their webhosting providers:
https://our.umbraco.org/forum/umbraco-7/using-umbraco-7/73445-umbraco-733-pngcmd

Can it be in a try/catch block so that people who don't have permission can at least load their images?

JimBobSquarePants:
By interesting you mean super clever yeah? :tongue: 

I'll stick it in a try catch but it's also possible to turn off the behaviour via configuration. 

http://imageprocessor.org/imageprocessor-web/configuration/#processingconfig

JimBobSquarePants:
I've been putting some thought to this and I don't want to mask an error in a try-catch. I want to encourage people to actually read the documentation on software they push.

samprogram:
I have this problem too! Took me hours to find this thread.. Please reconsider, people are not going to magically discover documentation for a dependency that Umbraco has and most people don't even know exists. I just saw that my images didn't work, nothing in the Umbraco logs.. :/

JimBobSquarePants:
Magically discover? There's been several blogpost on the Umbraco website, it's listed as a dependency, there's a Uhangout episode, I've handwritten comprehensive documentation and I've even done talks about it.... There's plenty of information out there.

I've been writing the code and documentation on this project, mostly in absolute isolation, for free, for the last 4 years. I'm tired of it.

I do not want to devote any more time to this version. I have my own priorities, my own life to lead. If you want the change I encourage you to fork the project and submit a pull request.

samprogram:
That reminds me of this talk which is enjoyed watching: (from "Open Source & Feelings"): https://www.youtube.com/watch?v=U15nkcSY2GI

I hope you feel better tomorrow and appreciate all your hard work on this project.

Thanks @nul800sebastiaan for the PR, hope it can be merged in and released soon.

JimBobSquarePants:
Fixed

PR #272, opened 2015-12-09
jpegtran errors with: "This program is blocked by group policy." #268

nul800sebastiaan:
Adds try-catch for this specific exception


JimBobSquarePants:
Thanks for that. Much appreciated.

Issue #273, opened 2015-12-14
Resizing Image with single dimension specified - not working as expected


thecodefish:
Hi, I'm trying to resize an image given a single, fixed dimension (ie 300 pixels wide). I want the image to keep the same aspect ratio, which means that the image will be a variable height after the resize operation. From the current documentation this is supported via the `ImageFactory.Resize(ResizeLayer resizeLayer)` method (http://imageprocessor.org/imageprocessor/imagefactory/resize/) - "Passing a single dimension will automatically preserve the aspect ratio of the original image".

However when I specify width = 300 and height = 0, I get an empty image returned. Similarly if I swap the dimensions around. However if I specify both width and height then it works as expected.

Of interest is that ImageProcessor returns me a 300x391 pixel image (which is expected, given the original image dimensions) - but the image is blank. In that case I was using the BoxPad resize mode.

I've had a very brief look at the code but couldn't tell much, other than I couldn't find a test for this scenario. If I can find the time I'll look into it in more detail (and maybe a PR if I figure things out) but I thought I'd create this issue anyway in case there's a different way of achieving what I want (or a known issue).

I searched for issues related to this and found #243 but that just adds the documentation I'm relying on.

I'm using NuGet package ImageProcessor 2.3.10

thecodefish:
I extended the unit tests that I found (https://github.com/JimBobSquarePants/ImageProcessor/blob/V2/src/ImageProcessor.UnitTests/ImageFactoryUnitTests.cs#L779) to perform both 'Pad' and 'BoxPad' operations when only a single dimension is specified. 

Here's my result for Pad with height=300
![resize-padsingledimension-height-0](https://cloud.githubusercontent.com/assets/2665874/11797542/871a49f8-a328-11e5-9462-c787c004cc8e.jpg)

Here's my result for BoxPad with height=300
![resize-boxpadsingledimension-height-0](https://cloud.githubusercontent.com/assets/2665874/11797573/c5bd2acc-a328-11e5-8a86-c76d8736f090.jpg)

JimBobSquarePants:
Hi there, Thanks for providing so much detail with the issue. It's very much appreciated!

I'll have a look as soon as I can to see what logic has been missed. That method has grown arms and legs over the years so it can be a shade tricky to navigate.

thecodefish:
[...]

PR #274, opened 2015-12-15
Fix for Resizing against one dimension with BoxPad

thecodefish:
Fixes issue #273 where Resizing with only one dimension specified in BoxPad
mode would result in a blank image


JimBobSquarePants:
Looks good. Will merge this evening when I get a chance to properly test. Much appreciated! :+1:

PR #285, opened 2016-01-02
Adding simple decode for jpeg grayscale colorspaces. Also added test …

rubensr:
This is a simple change to allow reading of Jpegs using grayscale colorspaces. I also added a new unit test image for that. 
It basically only affects the conversion of the input Jpeg image object into the RGB components - I decided to split it up into an "if" statement but feel free to ignore all this and write a more concise solution.


JimBobSquarePants:
This is great stuff. I think the if statement is most probably the best way to go about it.

Thanks!

PR #288, opened 2016-01-03
Simple change to emulate the V2 resize behaviour - if either width or…

rubensr:
… height is zero, then the image is scaled keeping its aspect ratio.
I've added a couple of simple tests for the new code.

Please let me know if you'd like me to change anything - these are my first pull requests in GitHub, so my code may not be up to the standards you're expecting, or you may have had other expectations in terms of where the tests should have been placed, etc.


JimBobSquarePants:
I kept thinking to add that behaviour. Thanks!

Issue #260, opened 2015-12-01
Add additional color spaces.


JimBobSquarePants:
More a nice to have than anything but I believe that if we are to do this properly we shouldbe able to  convert where possible between different color spaces.

Currently we have the following.
- [x] Color - Float based, premultiplied alpha, No limit to r, g, b, a values allowing for a fuller color range.
- [x] BGRA32
- [x] CIE Lab
- [x] CIE XYZ
- [x] CMYK
- [x] HSV
- [x] HSL
- [x] YCbCr

Inspiration could be taken from the following:

http://colormine.org/
https://github.com/scijs/color-space

christopherbauer:
Looks like we have the two missing color spaces merged into the v3 branch.

JimBobSquarePants:
Yup... Forgot to close this. Thanks!

tompazourek:
I wrote a .NET library that deals with color spaces. Might be useful:

https://github.com/tompazourek/Colourful

JimBobSquarePants:
Hi Tom,

I'm sure that would be very useful. Would it be possible for you to cast your eye over the work we have done already to see if we have done anything daft?

Cheers

James

tompazourek:
Hi James,

I had a brief look at the color conversions. Just few observations and notes:

- Note that in the RGB color space the color is unambiguous only if accompanied by an [RGB color space (working space)](https://en.wikipedia.org/wiki/RGB_color_space). For, example this means that when you have an RGB triplet (0.6, 0.7, 0.1), it will represent slightly different colors if you say that the color space is sRGB or other (Apple RGB, Adobe RGB, CIE RGB). But I guess that **for simplicity, you can just assume that all RGB triplets are in sRGB**. Each of these RGB color spaces can have different reference colors, reference white points and different mechanisms for compressing the channels from their linear versions.
[...]

PR #290, opened 2016-01-05
Colorspace CIE Lab

rubensr:
Hi Jim,

 Here is a branch that contains the CIE Lab color space and the tests that go with it, to go with #260.
 Hopefully nobody else was already working on this. I hope I properly followed your suggestions posted in regards to the CIE xyz space, but since the RGB conversions for Lab seem to go through xyz, you may want my branch refactored to use the XYZ functions once those are in. In any case, let me know. 

The test values I used were matched against ColorMine.

### DIFF
--- a/src/ImageProcessor/Colors/ColorTransforms.cs
+++ b/src/ImageProcessor/Colors/ColorTransforms.cs
@@ -192,11 +192,16 @@ namespace ImageProcessor
             float y3 = y * y * y;
             float z3 = z * z * z;
 
-            y = (y3 > 0.008856F) ? y3 : (y - 16F / 116F) / 7.787F;
             x = (x3 > 0.008856F) ? x3 : (x - 16F / 116F) / 7.787F;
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
This is excellent. Very much appreciated! :+1: 

I'm impressed with the work done to maintain accuracy. I need to check over some of the other color tests to ensure they are as well done.

I think the best bet would be to merge both this pull request and the XYZ one in its current state and continue on from there. If I merge them would you be able to continue?

rubensr:
Sure, of course - no problem.

dampee:
Littlebit busy ATM to continue. Ill have more time in a few weeks.

Op 5-jan.-2016 11:42 PM schreef rubensr <notifications@github.com>:

Sure, of course - no problem.

-

JimBobSquarePants:
Ace, I'll merge them in now then. Thanks both of you.

PR #291, opened 2016-01-06
XYZ code updated, tests working against ColorMine.org.

rubensr:
Hi Jim, here is the updated XYZ code + tests that match ColorMine.

However I noticed a couple of things that I did not know how you wanted to handle:

There are tests I copied that try to generate random RGB colours and then match them against the converted XYZ, Lab, Hsv, etc. I believe there are a couple of issues here:

1) Code is using Random.next(1) to create the values, which returns an integer only, not a number between 0 and 1;
2) Once I corrected the above, then I started to come across problems with the different implementations of Equals() for the Color Space structs - some try to perform exact matches using Vector4.Equals() (which fails a lot of the time due to floating point calculations), some do Equals() using epsilons.

So I was not sure how you wanted to deal with this, so did not touch any of it.

### DIFF
--- a/src/ImageProcessor/Colors/ColorTransforms.cs
+++ b/src/ImageProcessor/Colors/ColorTransforms.cs
@@ -67,29 +67,27 @@ namespace ImageProcessor
             return new Color(r, g, b);
         }
 
+        /// <summary>
+        /// Allows the implicit conversion of an instance of <see cref="CieXyz"/> to a
+        /// <see cref="Color"/>.
+        /// </summary>
+        /// <param name="color">The instance of <see cref="CieXyz"/> to convert.</param>
+        /// <returns>
+        /// An instance of <see cref="Color"/>.
+        /// </returns>
         public static implicit operator Color(CieXyz color)
         {
-            var x = color.X;
-            var y = color.Y;
-            var z = color.Z;
+            float x = color.X / 100F;
+            float y = color.Y / 100F;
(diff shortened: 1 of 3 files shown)


JimBobSquarePants:
Thanks for that! :+1: 

Yay unit tests!  That's interesting to discover... I best start investigating what is happening. Equals should be consistent across structs. I'm up for any suggestions.

rubensr:
Hmm I am probably not an expert in the field to weigh in with a lot of knowledge to back me up, but by  looking at the ColorMine source [code](https://github.com/THEjoezack/ColorMine/blob/master/ColorMine/ColorSpaces/Comparisons/Cie1976Comparison.cs) and the article that it [points to](https://en.wikipedia.org/wiki/Color_difference#CIE76), this again seems to be more complex than it looks :)

In short after reading all that I am not sure if our "Equals()" should answer the question "are these two colours so close that we can't distinguish them" or if it should only answer "are the RGB/XYZ/LAB components close enough based on some hand-picked epsilon".

JimBobSquarePants:
I ended up creating an interface `IAlmostEquatable<T, TP>` that the various colorspaces implemented that allowed precision based epsilon. This made it less prone to false errors. 

I have commented out the random parts of the tests for now though as I'm not sure that they are a good test. There will always be a loss of accuracy when converting to/from the colorspaces depending on the processor architecture.

Issue #294, opened 2016-01-07
Blend modes


JimBobSquarePants:
It would be great to have the ability to blend colors based on the W3c specification.

http://www.w3.org/TR/compositing-1/#blending

Additional blend modes can be added as static methods on ColorSpacialTransforms in ColorSpacialTransforms.cs

Blend Modes:
- [ ] Separable blend modes
  - [ ] normal blend mode
  - [x] multiply blend mode
  - [ ] screen blend mode
  - [ ] overlay blend mode
  - [ ] darken blend mode (committed awaiting review)
  - [ ] lighten blend mode (committed awaiting review)
  - [ ] color-dodge blend mode
  - [ ] color-burn blend mode
  - [ ] hard-light blend mode
  - [ ] soft-light blend mode
  - [ ] difference blend mode
  - [ ] exclusion blend mode
- [ ] Non-separable blend modes
  - [ ] hue blend mode
  - [ ] saturation blend mode
  - [ ] color blend mode
  - [ ] luminosity blend mode

christopherbauer:
Hey Jim, I was eyeing for something like this, but my initial thought seemed like this might be too simple. I don't have internet atm but the pseudo code I came up with was:

 Public color multiply (color destination)
    If(destination == Black)
        Return black
    If(destination == White)
         Return this
    Return new Color(this.r * destination.r, this.g * destination.g, this.B * destination.B)

Could it be so simple? If so when I get home I'll start right on it with tests happily.

christopherbauer:
All my whitespace was removed! Damn

JimBobSquarePants:
Hi Christopher,

The only complexity involved that I see is that the color component values in the Color class are premultiplied. I'm not sure what issues that would cause with the formulas given.

christopherbauer:
I had to laugh - I was double checking the link and read the line "... In practice, this is usually all performed in one step. The blending calculations must not use pre-multiplied color values." D'oh!

JimBobSquarePants:
Aye... Now we have Maths to deal with. I wish I'd paid attention more in school!

christopherbauer:
Took a crack at it in my branch, just as a pre-multiplied basic version. What I plan to do is if this appears to be in the right direction I'll investigate using some of the library functions to convert it out to post-multiplied and perform the multiply blending to the post-multiplied color.

Could you take a look and give some feedback if I'm moving in the right direction?

[...]

PR #298, opened 2016-01-13
Multiply blending

christopherbauer:
This is a branch that satisfies some base criteria for multiply blending per issue https://github.com/JimBobSquarePants/ImageProcessor/issues/294. Discussion re: the implementation of multiply blending can continue here as work moves forward while generic information about blending can occur in the issue.

Or if this is okay as is we can merge in and investigate other modes.

### DIFF
--- a/src/ImageProcessor/Colors/ColorSpacialTransforms.cs
+++ b/src/ImageProcessor/Colors/ColorSpacialTransforms.cs
@@ -1,0 +1,31 @@
+﻿// <copyright file="ColorSpacialTransforms.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageProcessor
+{
+    using System.Numerics;
+
+    public partial struct Color
+    {
+        public static Color Multiply(Color source, Color destination)
+        {
+            if (destination == Color.Black)
+            {
+                return Color.Black;
+            }
+            if (destination == Color.White)
+            {
+                return source;
+            }
+            return
+                new Color(
+                    new Vector4(
+                        source.backingVector.X * destination.backingVector.X,
+                        source.backingVector.Y * destination.backingVector.Y,
+                        source.backingVector.Z * destination.backingVector.Z,
+                        source.backingVector.W * destination.backingVector.W));
+        }
+    }
+}
--- a/tests/ImageProcessor.Tests/Colors/ColorSpacialTransformTests.cs
+++ b/tests/ImageProcessor.Tests/Colors/ColorSpacialTransformTests.cs
@@ -1,0 +1,53 @@
+namespace ImageProcessor.Tests
+{
+    using Xunit;
+
+    public class ColorSpacialTransformTests
+    {
+        public class MultiplyTests
+        {
+            [Fact]
+            public void MultiplyBlendConvertsRedBackdropAndGreenOverlayToBlack()
+            {
+                var backdrop = Color.Red;
+                var overlay = Color.Green;
+
+                var result = Color.Multiply(backdrop, overlay);
+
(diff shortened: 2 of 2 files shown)


christopherbauer:
RE: The spec to black/white: multiplying by black would be multiplying R G and B with 0 which would return black anyway, so this is more of a 'skip the cpu cycles if it is black, we don't really need to bother. Same thing for white - multiplying by 1f across the board results in the same value as you started. Though I suppose the implementation of equals in the lib will incur more cycles then just performing the multiplication outright, huh? :)

JimBobSquarePants:
I've not tested that actually but I would imagine it would do since I don't actually use as strict equals in my equality operators, rather compare to an epsilon.

JimBobSquarePants:
This is my fault for giving you a name that already exists but could you separate out the colorspace transforms into a separate file, they make Color.cs difficult to traverse. Call it ColorSpacialTransforms

christopherbauer:
Hey Jim,

That was just a weird Github merge error in the diffs. I don't know why it appears to be moving Color out of the file ColorTransform.cs but rest assured your sanity is in check, the file did not exist prior to the branch :)

JimBobSquarePants:
Nah something's wrong here chum. A quick dig in your repo yields 

https://github.com/christopherbauer/ImageProcessor/blob/MultiplyBlending/src/ImageProcessor/Colors/Color.cs

With the colorspace translations present. In my V3 branch you will find a file ColorTransforms.cs that contains those methods.

christopherbauer:
OH! It's a file but the file contains a public partial Color declaration. I'm used to files being named after the classes that contain them and completely missed it. I must have moved the multiply declaration into ColorTransform.cs then alt+entered the partial color struct declaration. I'll clean up the commit ASAP.

JimBobSquarePants:
No worries :smile: Yeah, taking a slightly different approach with that struct. There's far too much to go in one file.

christopherbauer:
I just squashed my commit tree down to a single commit for brevity's sake and to keep from polluting the branch with my bad commits :+1:

src/ImageProcessor/Colors/ColorSpacialTransforms.cs
```diff
@@ -0,0 +1,26 @@
+﻿// <copyright file="ColorSpacialTransforms.cs" company="James Jackson-South">
+// Copyright (c) James Jackson-South and contributors.
+// Licensed under the Apache License, Version 2.0.
+// </copyright>
+
+namespace ImageProcessor
+{
+    public struct ColorSpacialTransforms
```
JimBobSquarePants: Should be `public partial struc Color`

src/ImageProcessor/Colors/ColorSpacialTransforms.cs
```diff
@@ -0,0 +1,26 @@
...
+            if (destination == Color.Black)
+            {
+                return Color.Black;
+            }
+            if (destination == Color.White)
+            {
+                return source;
+            }
+            return new Color(
+                (source.R * destination.R).Clamp(0f, 1f),
```
JimBobSquarePants: Once you use the correct partial you can then multiply the backing vectors. Don't limit as I want to support HDR

src/ImageProcessor/Colors/ColorSpacialTransforms.cs
christopherbauer: I'm falling to pieces!

JimBobSquarePants: Tell me about it! I'm banging my head against a brick wall just now trying to speed up the resampler.

christopherbauer:
Okay - working with the backing vector, corrected the struct issue, merged from origin, and squashed the commits so it's not messy!

JimBobSquarePants:
perfick

PR #301, opened 2016-01-15
When WebP quality is set to 100, encode losslessly.

optlink:


JimBobSquarePants:
Looks perfect! Thanks! :100:

Issue #303, opened 2016-01-18
ImageProcessor.Web.AzureBlobCache.1.0.6 sets incorrect Content Type


perosb:
In the latest 4.5.0 version it looks like Content-Type is default set to "text/html".

This seem to be for the second request when coming from cache. (in my case Azure Blob cache)

The first request set proper image/* header.

Prior to 4.5.0, the second request doesn't set any header at all, hence it still seem to work.

perosb:
This commit looks suspecious: https://github.com/JimBobSquarePants/ImageProcessor/commit/2f6e9f8c04c0f843b2e00f748e66f6d8550d0b24

perosb:
https://github.com/JimBobSquarePants/ImageProcessor/issues/276

JimBobSquarePants:
Hmmm... `this.mimetype` is not null then that means the content type of the blob should have been set. 

https://github.com/JimBobSquarePants/ImageProcessor/blob/2f6e9f8c04c0f843b2e00f748e66f6d8550d0b24/src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs#L223

If it is this causing the problem then it's not ImageProcessor.Web but the AzureBlobCache setting the wrong type.

perosb:
It might be better to default to an unknown content type, image/xyz and have the browser decide.
Right not, text/html is not correct.

perosb:
From what I can see, the Content-Type header is set correctly on the Blob.
Also looks ok, when setting the metadata.

perosb:
@JimBobSquarePants think I fixed it.
issue was that using this.mimetype which is only set when adding to cache instead of using the GET requests headers.

perosb:
Also properly fixes https://github.com/JimBobSquarePants/ImageProcessor/issues/276 I believe

perosb:
Let me know if there is anything more i can do here.
I believe the PR https://github.com/JimBobSquarePants/ImageProcessor/pull/304 fixes the issue currently breaking all AzureBlobCache images.

JimBobSquarePants:
Sorry, I'll get this merged in. I'm juggling a few too many things just now.

perosb:
No worries @JimBobSquarePants 
Just saw 4.5.1 and thought there still were issues with the PR.

Thanks!

JimBobSquarePants:
[...]

PR #304, opened 2016-01-18
set unknown content type (but still image) as fallback when unknown

perosb:
workaround for https://github.com/JimBobSquarePants/ImageProcessor/issues/303

### DIFF
--- a/src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
+++ b/src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
@@ -381,11 +381,12 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
                         {
                             HttpResponse contextResponse = context.Response;
                             cachedStream.CopyTo(contextResponse.OutputStream);
+                            string contentType = response.ContentType;
 
                             // Mimetype can be null when returning from the cache.
                             ImageProcessingModule.SetHeaders(
                                 context,
-                                string.IsNullOrWhiteSpace(this.mimeType) ? contextResponse.ContentType : this.mimeType,
+                                string.IsNullOrWhiteSpace(contentType) ? "image/xyz" : contentType,
                                 null,
                                 this.MaxDays,
                                 response.StatusCode);


src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -381,11 +381,12 @@ public override void RewritePath(HttpContext context)
                         {
                             HttpResponse contextResponse = context.Response;
                             cachedStream.CopyTo(contextResponse.OutputStream);
+                            string contentType = response.ContentType;
 
                             // Mimetype can be null when returning from the cache.
                             ImageProcessingModule.SetHeaders(
                                 context,
-                                string.IsNullOrWhiteSpace(this.mimeType) ? contextResponse.ContentType : this.mimeType,
```
JimBobSquarePants: There must be a way we can guarantee the correct type? I go to great pains to ensure that the correct type is sent to Azure.

perosb:
@JimBobSquarePants yes, I just updated to PR with the correct fix.

src/Plugins/ImageProcessor.Web/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
perosb: PR has been updated with correct fix.

JimBobSquarePants: You sure? It still says `"image/xyz"` in the PR.

perosb: @JimBobSquarePants  yes, proper fix is to get ContentType from response/azure.

the "image/xyz" is still there, but is just a fallback if Azure doesn't.

perosb: I mean, the proper fix _is_ included in the PR.

PR #306, opened 2016-01-18
Fix #305, broken float parsing on non en-US cultures

### LINKED ISSUE #305: Version 4.5.0 breaks percentage cropping
When using a predefined crop in Umbraco, it generates a URL like this: http://localhost:1823/media/1019/loebere-paa-strand.jpg?crop=0.020427188497102129,0,0.72831839298854173,0.6610013498588786&cropmode=percentage&width=246&height=221&rnd=130975950450000000

This works fine with ImageCropper.Web 4.4, but breaks with 4.5. It throws an ArgumentException in the Crop processor. And debugging, it is quite understandable why: 

0.020427188497102129 in the URL somehow becomes 2.04271887E+16. The bug seems to lie in the class GenericConvertableConverter, on this line: 

    return (T)Convert.ChangeType(value, t);

In short, it seems there has been some change regarding the parsing of float values, as that is where the issue comes from. In particular, commit 189908982cd1ddcec7bccd76d99d0266403779ca seems to be the problematic one.

**Steps to reproduce (using Umbraco)**

1. Update your Umbraco site to ImageProcessor.Web 4.5
2. Set your Umbraco site language to da-DK (or any culture that uses comma as decimal separator) in Settings, Languages (just changed default en-US to da-DK).
3. Add a preset crop to Image Cropper datatype
4. Upload an image. 
5. Go to the image and customize the crop by clicking it and dragging it around and zooming in.
6. Then try to render the named crop using GetCropUrl. You will get an HTTP 500 error instead of the image.

arknu:
The culture was not passed to Convert.ChangeType, which caused issues with float parsing when running on cultures where comma is the decimal separator. The culture parameter (which is correctly passed as InvariantCulture ensures that the correct parsing rules are used.

### DIFF
--- a/src/ImageProcessor.Web/Helpers/QuerystringParser/Converters/GenericConvertableConverter.cs
+++ b/src/ImageProcessor.Web/Helpers/QuerystringParser/Converters/GenericConvertableConverter.cs
@@ -49,7 +49,7 @@ namespace ImageProcessor.Web.Helpers
                     return (value == null) ? default(T) : (T)Convert.ChangeType(value, u);
                 }
 
-                return (T)Convert.ChangeType(value, t);
+                return (T)Convert.ChangeType(value, t, culture);
             }
 
             return base.ConvertFrom(culture, value, propertyType);


JimBobSquarePants:
Annoyed at myself a little for missing this. Thanks!

kalevigit:
I have tried switching 
CultureInfo.DefaultThreadCurrentCulture = new CultureInfo("en-US");

in Application_Start, but its not helping.. what am I to do?? :(

arknu:
@kalevigit This is not really the place to discuss it, but upgrade to ImageProcessor 4.5.1 which has this fix described here. If that doesn't help, it is a different issue and you should file a new issue with steps to reproduce it.

Issue #308, opened 2016-01-20
OutOfMemory exception processing animated GIF


nul800sebastiaan:
Spent an hour to try to find some workarounds to fix an OutOfMemory exception when processing the GIF below. There error occurs after processing about 280 out of th 675 frames on this line of code: 
https://github.com/JimBobSquarePants/ImageProcessor/blob/V2/src/ImageProcessor/Imaging/Formats/GifDecoder.cs#L59

The error was reported here: http://issues.umbraco.org/issue/U4-7720 as "The parameter is Invalid" but when I run it, it's consistently crashing on an OutOfMemory.

Instead of keep trying to Google this and finding workarounds that make things worse I thought I'd ask the expert first to see if he knows a better tactic of the top of his head! ;-) Any ideas? Happy to work on this some more once I have a direction to try.

![ogb-insider-blogs-googlelogox2-animated](https://cloud.githubusercontent.com/assets/304656/12455610/ac271448-bf9c-11e5-8e90-8215cdc178ed.gif)

JimBobSquarePants:
Hey chum,

There's no way I can think of that you'll get this to work with the present codebase - there's just too many frames.

Here's essentially what is happening. For `System.Drawing` to be able to edit an individual image frame it needs to be able to load it expanded into memory. With the way it's written just now that means we are attempting to load 675 images, each at 1280x720. We're successfully loading 280 concurrent images at the same time which I think is pretty good!

There's a big part of me that simply says "Use Photoshop"  :tongue: for images like this as I never thought someone would attempt to edit a gif with so many frames but there might be a way with some refactoring to vastly reduce the amount of memory used.

In the `GifDecoder` we create all the frames at once then add them to the encoder in a loop.

What we could do instead is create each frame on demand, process it, add it to the `GifEncoder` then dispose of it.  By adding a method to the `GifDecoder` that allows you to access each frame by index it should be possible... All the information is there to add to the `GifEncoder` without having to create hundreds of images.

I *might* have a chance to look at it today but if you want to have a crack, go for it and see how you get on. 

I might just work!

JimBobSquarePants:
Thinking about v3 I extract all the frames on load. Time to rethink that process!

nul800sebastiaan:
Thanks for the feedback! I may have some time to look into it today but I'm teaching courses, so I'm fully occupied until the end of the day. If they don't wear me out too much today I can have a look in the evening.

coolmikkel:
[...]

PR #322, opened 2016-02-03
OutOfMemory exception processing animated GIF #308

nul800sebastiaan:
Learned some stuff about the code and noticed that what's handled in `GifDecoder` gets sent to `ImageExtensions.Copy`, sweet! So only need to update code in one place (I should've known).

So, this is just starting the conversation and explains a bit of what I've tried:
- I tried taking 10 frames from that giant 694 frame GIF and making an animation out of that with disastrous results (so this took frame number 0, 69, 69x2, 69x3, etc until 69x10 and mashed them together): 
  
ogb-insider-blogs-googlelogox2-animated 1
- I don't think processing many more than 10 frames is a very good idea, if there's many gifs, it will be rough on CPU/memory. 
- Decided to stick with just showing the first frame
  - This is currently hardcoded but needs to be configurable from querystring (please advise, not sure how to get the variable in the `GifDecoder`)
- The code is currently doing: 

```
                        // Reset the position.
                        if (i == last)
                        {
                            image.SelectActiveFrame(FrameDimension.Time, 0);
                        }
```

I don't know what this achieves, but if this is required to run then I need to fix it here for when there's only one frame processed (please advise).

### DIFF
--- a/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
+++ b/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
@@ -12,6 +12,7 @@ namespace ImageProcessor.Web.HttpModules
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.Specialized;
     using System.IO;
     using System.Linq;
     using System.Net;
@@ -21,6 +22,7 @@ namespace ImageProcessor.Web.HttpModules
     using System.Web;
     using System.Web.Hosting;
 
+    using ImageProcessor.Imaging;
     using ImageProcessor.Imaging.Formats;
     using ImageProcessor.Web.Caching;
     using ImageProcessor.Web.Configuration;
@@ -80,7 +82,7 @@ namespace ImageProcessor.Web.HttpModules
         /// without querystring parameters.
         /// </summary>
         private static bool? interceptAllRequests;
-
+        
         /// <summary>
         /// A value indicating whether this instance of the given entity has been disposed.
         /// </summary>
@@ -529,7 +531,9 @@ namespace ImageProcessor.Web.HttpModules
                         // Process the image.
                         bool exif = preserveExifMetaData != null && preserveExifMetaData.Value;
                         bool gamma = fixGamma != null && fixGamma.Value;
-                        using (ImageFactory imageFactory = new ImageFactory(exif, gamma))
+                        AnimationProcessMode mode = this.ParseAnimationMode(queryString);
+
+                        using (ImageFactory imageFactory = new ImageFactory(exif, gamma) { AnimationProcessMode = mode })
                         {
                             byte[] imageBuffer = null;
                             string mimeType;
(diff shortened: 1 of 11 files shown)


JimBobSquarePants:
Good to see progress made chum!

I think we'd have to add an interface to the `GifFormat` containing the `AnimationProcessMode` enum, `IAnimatedImageFormat`. Doing that should allow us to check for the interface and assign the value by adding the same enum as a property on the `ImageFactory` class. By adding the interface we can keep a minor version also since we are not adding a property to `ISupportedImageFormat`

In the `ImageFactory.Load()` variants 

``` c#
if (this.CurrentImageFormat is IAnimatedImageFormat){
    ((IAnimatedImageFormat)this.CurrentImageFormat).AnimationProcessMode = this.AnimationProcessMode;
}
```

In the HttpModule in ImageProcessor.Web we could detect the querystring value using the `QueryParamParser` which can handle enums.  Using that you could set the factory property. It would have to be done separately from any standard `IWebGraphicsProcessor` since it needs to happen on initialization. You want to do something like.

``` c#
AnimationProcessMode mode = this.ParseAnimationMode(queryString);

using (ImageFactory imageFactory = new ImageFactory(exif, gamma){ AnimationProcessMode = mode })
{
    // Do the normal stuff
}
```

As for this bit of code:

``` c#
 // Reset the position.
if (i == last)
{
    image.SelectActiveFrame(FrameDimension.Time, 0);
}
```

That resets the animated image so that we are reading from the correct first frame when performing additional processing. 

Does that make sense to you?

nul800sebastiaan:
Alright, all updated as per your hints. 

One thing I really dislike here is that I had to alter the `Copy` method on `ImageExtensions` to accept an `AnimationProcessMode`. I have a feeling that this could be done better.
The thing here that raised a red flag is: I'm already setting the mode in the `ImageFactory` only to then immediately pass that mode into the `Copy` method. I also couldn't find a way to pass the `AnimationProcessMode` from `Resizer` into the linear resizer, which also calls this `Copy` method. Might not be a concern but it all feel a little dirty to me. :) It would be much nicer if `GifDecoder` knew about the factory but then again that may be crossing other lines. :)

Alright, I understand setting the image to the first frame now, but I feel it should be the responsibility of the `Copy` method then, I may have missed something? I have moved it for now but that might not be correct, let me know!

Look forward to your feedback!

JimBobSquarePants:
The code so far lloks good. It think we're restricted simply because the library was never designed for this. I mean, the whole GifDecoder/Encoder thing is a big hack anyway really.

That Copy method could be an issue though. Are we disappearing down a rabbit hole?

nul800sebastiaan:
Neh, I think it's all pretty solid as long as we don't have to tack on 18 more of these options ;-)

I'll make corrections as suggested, thanks!

nul800sebastiaan:
Oh actually, ignore that last commit, it doesn't even build, silly me! Will update tomorrow.

JimBobSquarePants:
Tut tut! :smile: No worries. It's good to see someone tinkering around.

nul800sebastiaan:
Silly me, all fixed now!

nul800sebastiaan:
I can squash these commits if you think this is all good, then we don't pollute the repo with my ineptitude ;)

nul800sebastiaan:
Very strange, by the way, that AppVeyor was able to build that broken code in the previous commit, VS wouldn't let me build..

JimBobSquarePants:
Aye odd isn't it! Yeah, go ahead. Just the Copy method in Resizer now to go isn't it? Or does it need editing since the gif should only have one frame when set on load anyway?

nul800sebastiaan:
Totally forgot about the Resizer, that's now added and it works too! Squashed to hide the shame of my mistakes along the way. ;-)

JimBobSquarePants:
Excellent! I'll merge this in and do some final testing before release.

nul800sebastiaan:
Whoop! It's my most ambitious PR yet! :-)

Now, who do I have to bribe to get it released today/tomorrow? Will send beer! (would love to include it in 7.4 final, due tomorrow-ish).

JimBobSquarePants:
I'll try and get it out tonight. Hold tight!

PR #326, opened 2016-02-11
Fix Watermark text centering

khiemtong:
I discovered an issue while Watermark where centering the text was always a bit off. It was more obvious when the string barely fits in the parent image.

I searched around and found the issue discussed here.

http://stackoverflow.com/questions/1203087/why-is-graphics-measurestring-returning-a-higher-than-expected-number

I've made two small changes to implement a solution. I've tested it locally against my own project and the changes look to be safe. If you need me to roll back the changes to `src/ImageProcessor/ImageProcessor.csproj` and `src/TestWebsites/MVC/Test_Website_MVC.csproj`, let me know, not sure what differences in VS caused them.


JimBobSquarePants:
Hi there, thanks for this. 

Could you please clean this up a little and rebase from the source though. There's only really one file changed so there's shouldn't be all these commit messages. 

A couple of screenshots showing the difference would be very useful also if you can manage that?

Cheers! :+1:

khiemtong:
I created an updated PR with screenshots and fixed the issue with commit messages, closing this PR

Please see https://github.com/JimBobSquarePants/ImageProcessor/pull/328

JimBobSquarePants:
Brilliant. Thanks!

JimBobSquarePants:
Ah yes. That's strikingly better. I'm gonna conduct some tests against vertical text then get this merged.

PR #328, opened 2016-02-12
Fix watermark text centering

khiemtong:
Updated PR from https://github.com/JimBobSquarePants/ImageProcessor/pull/326

> I discovered an issue while Watermark where centering the text was always a bit off. It was more obvious when the string barely fits in the parent image.

>I searched around and found the issue discussed here.

>http://stackoverflow.com/questions/1203087/why-is-graphics-measurestring-returning-a-higher-than-expected-number

Test code

```csharp
			using (var outStream = new MemoryStream())
			using (var imageFactory = new ImageFactory(false))
			{

				imageFactory.Load(imgBytes)
					.Watermark(new TextLayer { Text = "This should be a long enough string - that the text is not center aligned", FontColor = Color.White, FontSize = 36 })
					.Save(outStream);

				File.WriteAllBytes("before-pr.png", outStream.ToArray());

			}
```

**Before**

before-pr

**After**

after-pr

### DIFF
--- a/src/ImageProcessor/ImageProcessor.csproj
+++ b/src/ImageProcessor/ImageProcessor.csproj
@@ -140,6 +140,7 @@
     <Compile Include="Configuration\NativeMethods.cs" />
     <Compile Include="Imaging\ComputerArchitectureInfo.cs" />
     <Compile Include="Imaging\AnimationProcessMode.cs" />
+    <Compile Include="Imaging\ComputerArchitectureInfo.cs" />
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Ah yes. That's strikingly better. I'm gonna conduct some tests against vertical text then get this merged

JimBobSquarePants:
Hi there, Can you repoint this at the Framework branch so I can merge it. Thanks!

khiemtong:
Opened #332 to target Framework branch. Closing this one.

PR #329, opened 2016-02-14
Add lighten/darken blend modes

christopherbauer:
For Issue #294, lighten and darken blend modes + tests.

Unfortunately I still can't get the xunit test runners working, so I can't check my logic here. Any takers for running the unit tests? It was straightforward enough that I feel confident (famous last words).

### DIFF
--- a/src/ImageProcessorCore/Colors/ColorTransforms.cs
+++ b/src/ImageProcessorCore/Colors/ColorTransforms.cs
@@ -6,6 +6,7 @@
 namespace ImageProcessorCore
 {
     using System;
+    using System.Numerics;
 
     /// <summary>
     /// Represents a four-component color using red, green, blue, and alpha data. 
@@ -18,9 +19,58 @@ namespace ImageProcessorCore
     public partial struct Color
     {
         /// <summary>
+        /// Blends two colors by lightening.
+        /// </summary>
+        /// <remarks>
+        /// Selects the lighter of the backdrop and source colors.
+        /// The backdrop is replaced with the source where the backdrop is lighter; otherwise, it is left unchanged
+        /// Note: This behavior is on a channel by channel basis, i.e., this rule is applied to each of the 3 RGB color channels separately.
+        /// </remarks>
+        /// <param name="backdrop"></param>
+        /// <param name="source"></param>
+        /// <returns>
+        /// The <see cref="Color"/>.
+        /// </returns>
+        public static Color Lighten(Color backdrop, Color source)
+        {
+            var newVector = new Vector4
+                                {
+                                    W = Math.Max(backdrop.backingVector.W, source.backingVector.W),
+                                    X = Math.Max(backdrop.backingVector.X, source.backingVector.X),
+                                    Y = Math.Max(backdrop.backingVector.Y, source.backingVector.Y),
+                                };
+
+            return new Color(newVector);
+        }
+        /// <summary>
+        /// Blends two colors by darkening.
+        /// </summary>
+        /// <remarks>
+        /// Selects the darker of the backdrop and source colors.
+        /// The backdrop is replaced with the source where the backdrop is darker; otherwise, it is left unchanged
+        /// Note: This behavior is on a channel by channel basis, i.e., this rule is applied to each of the 3 RGB color channels separately.
+        /// </remarks>
+        /// <param name="backdrop"></param>
+        /// <param name="source"></param>
+        /// <returns>
+        /// The <see cref="Color"/>.
+        /// </returns>
+        public static Color Darken(Color backdrop, Color source)
+        {
+            var newVector = new Vector4
+                                {
+                                    W = Math.Min(backdrop.backingVector.W, source.backingVector.W),
+                                    X = Math.Min(backdrop.backingVector.X, source.backingVector.X),
+                                    Y = Math.Min(backdrop.backingVector.Y, source.backingVector.Y),
+                                };
+
+            return new Color(newVector);
+        }
+
+        /// <summary>
         /// Blends two colors by multiplication.
         /// <remarks>
-        /// The source color is multiplied by the destination color and replaces the destination.
+        /// The source color is multiplied by the destination color and replaces the source.
         /// The resultant color is always at least as dark as either the source or destination color.
         /// Multiplying any color with black results in black. Multiplying any color with white preserves the 
         /// original color.
--- a/src/ImageProcessorCore/Colors/ColorTransforms.cs
+++ b/src/ImageProcessorCore/Colors/ColorTransforms.cs
@@ -35,9 +35,10 @@ namespace ImageProcessorCore
         {
             var newVector = new Vector4
                                 {
-                                    W = Math.Max(backdrop.backingVector.W, source.backingVector.W),
+//                                    W = Math.Max(backdrop.backingVector.W, source.backingVector.W),
                                     X = Math.Max(backdrop.backingVector.X, source.backingVector.X),
                                     Y = Math.Max(backdrop.backingVector.Y, source.backingVector.Y),
+                                    Z = Math.Max(backdrop.backingVector.Z, source.backingVector.Z),
                                 };
 
             return new Color(newVector);
@@ -59,9 +60,10 @@ namespace ImageProcessorCore
         {
             var newVector = new Vector4
                                 {
-                                    W = Math.Min(backdrop.backingVector.W, source.backingVector.W),
+//                                    W = Math.Min(backdrop.backingVector.W, source.backingVector.W),
                                     X = Math.Min(backdrop.backingVector.X, source.backingVector.X),
                                     Y = Math.Min(backdrop.backingVector.Y, source.backingVector.Y),
+                                    Z = Math.Min(backdrop.backingVector.Z, source.backingVector.Z),
                                 };
 
             return new Color(newVector);
--- a/tests/ImageProcessorCore.Tests/Colors/ColorSpacialTransformTests.cs
+++ b/tests/ImageProcessorCore.Tests/Colors/ColorSpacialTransformTests.cs
@@ -4,6 +4,117 @@ namespace ImageProcessorCore.Tests
 
     public class ColorSpacialTransformTests
     {
+        public class DarkenTests
+        {
+            [Fact]
+            
+            public void DarkenBlendConvertsToDarkerRedColorChannel()
+            {
+                var darkerRed = 25;
+                var backdrop = new Color(50, 0, 0);
+                var source = new Color(darkerRed, 0, 0);
+
+                var result = Color.Darken(backdrop, source);
+
+                Assert.Equal(result.R, darkerRed);
+            }
+            [Fact]
+            
+            public void DarkenBlendConvertsTodarkerGreenColorChannel()
+            {
+                var darkerGreen = 100;
+                var backdrop = new Color(0, darkerGreen, 0);
+                var source = new Color(0, 130, 0);
+
+                var result = Color.Darken(backdrop, source);
+
+                Assert.Equal(result.G, darkerGreen);
+            }
+            [Fact]
+            
+            public void DarkenBlendConvertsTodarkerBlueColorChannel()
+            {
+                var darkerBlue = 8;
+                var backdrop = new Color(0, 0, darkerBlue);
+                var source = new Color(0, 0, 30);
+
+                var result = Color.Darken(backdrop, source);
+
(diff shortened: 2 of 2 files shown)


JimBobSquarePants:
I think the xunit/Resharper issue is due to xunit not doing any updates until Corefx hits RC2. I am able, however, to run all my tests via the Test Explorer window in Visual Studio 2015 without issue. 

vs

src/ImageProcessorCore/Colors/ColorTransforms.cs
```diff
@@ -18,9 +19,60 @@ namespace ImageProcessorCore
...
+        /// Note: This behavior is on a channel by channel basis, i.e., this rule is applied to each of the 3 RGB color channels separately.
+        /// </remarks>
+        /// <param name="backdrop"></param>
+        /// <param name="source"></param>
+        /// <returns>
+        /// The <see cref="Color"/>.
+        /// </returns>
+        public static Color Lighten(Color backdrop, Color source)
+        {
+            var newVector = new Vector4
```
JimBobSquarePants: What happens with the alpha channel here?

christopherbauer: I didn't want to work on alpha because I'm not sure how to handle the pre 
vs post situation. What is the intent of the transforms, to be used for 
internal image processor library operations or to be used by end users? If 
it's internal we can assume all colors are post and provide one 
implementation. If we have to support pre and post then we need to discuss 
how to architect a solution that can be used based on the source colors pre 
or post status.

JimBobSquarePants: Since they are static methods on the `Color` struct they would be used in either scenario. Looking at the spec I would say it should take the higher of the two alpha values.

christopherbauer: This is from the w3c spec:

The result of the mixing function is modulated by the backdrop alpha. A 
fully opaque backdrop allows the mixing function to be fully realized. A 
transparent backdrop will cause the final result to be a weighted average 
between the source color and mixed color with the weight controlled by the 
backdrop alpha. The value of the new color becomes:

Cr = (1 - αb) x Cs + αb x B(Cb, Cs)

with:

Cr: the result color
B: the formula that does the blending
Cs: the source color
Cb: the backdrop color
αb: the backdrop alpha

-----------------------------

Which is why I've been hemming and hawing about pre multiplied alphas. If 
we have already added the alpha as part of the color in the library than we 
would need to undo that to fit the spec because the blending function is 
modulated by the alpha before the blend is applied to the color.

JimBobSquarePants: We'll have to do the same as `Lerp`. Check whether the alpha component is less than 1 and choose an appropriate algorithm. My head hurts thinking of how to translate it though.

christopherbauer: :8ball: I must be misunderstanding something about pre/post-multiplied alpha. I was under the impression that for all intents and purposes it could be represented as: 

Pre-multiplied: R,G,B,A
Post-multiplied: (R*A),(G*A),(B*A),A
:.
Pre-multiplied red w/ 50% opacity: 255, 0, 0, .5
Post-multiplied red w/ 50% opacity: 127.5, 0, 0, .5

Where pre-mult is the 'true color' and post-mult is the 'color after multiplication by alpha'. Such that we keep the same alpha value in post and pre, but the RGB is different. We can convert from post to pre by correcting for the alpha, but if we don't know whether or not a color X is post or pre then we can't do anything without potentially effing it up.

JimBobSquarePants: >Pre-multiplied red w/ 50% opacity: 255, 0, 0, .5
>Post-multiplied red w/ 50% opacity: 127.5, 0, 0, .5

You're right about the calculation those values should be below in your example though. (255 is bytes in Bgra32)

    1, 0, 0, .5
    .5, 0, 0, .5

Is `Cr = (1 - αb) x Cs + αb x B(Cb, Cs)` the formula then that is applicable for both opaque and premultiplied values?

All the colours in the Image class are premultiplied when loaded from the various image formats so we should assume that everything is premultiplied.

christopherbauer: The `Cr = (1 - αb) x Cs + αb x B(Cb, Cs)` works for opaque pre-multpilied because you're multiplying the blending function by 1! :+1: 

Now we have to determine if we're intending the ColorExtensions to be ImageProcessor internal or to be used by clients too, because we'll need to differentiate a "post multiplied blend" with a "pre multiplied blend" in the API for client use. What do you think?

JimBobSquarePants:
I think the public will want to use them also so yeah we should probably offer both variants.

christopherbauer:
Ok cool I'll get to fixing the implementation here and only use the 'pre' version for internal stuff. I'm sorry it was rather roundabout to get to the core of the issue but I'm glad we figured it out :+1:

JimBobSquarePants:
Don't be sorry chum, Discussion about it is great, we're both learning this stuff as we go along. :smile:

christopherbauer:
Ok, I've been mulling it over. We have a color object but ATM it doesn't track what the current status it is - it just 'is' a color and we have no way of marking it pre-multipied or not. Even though we know our colors are premultiplied, the users need to have an option to 'fix' their color.

Options - 

1. "Right Way" - Add a default-valued bool 'IsPreMultiplied' to the constructor of Color (using telescoping constructor pattern if necessary). We can default it to the scenario most users will require, then the implementations of blend modes and such can just simply use the correct implementation of the algorithms depending on circumstance. It also gives us a really simple pattern for identifying blend mode color errors - 
Client - "Hey the color is slightly off." 
AUSD (Abused Unpaid Support Dev :)): "Try switching to post-multiplied in the constructor!" 
Client - <disappears in the night never to be heard from again>
2. Ghetto Easy - Provide a PreMultiplyX + PostMultiplyX and let the user decide which their color is. This just seems like an awful experience for the end user, who would think that there method by which we perform lighten wouldn't be intelligent enough to know? 
3. Overkill Painful - Make color itself and abstract and have two derived color classes - PreMultipiedColor + PostMulipliedColor - then provide either a static class on the base color that changes the implementation based on type (oh god no) or just provide static implementations on each derived class. I hate this option don't pick it.

What do you think is the best?

JimBobSquarePants:
Options -

> "Right Way" - Add a default-valued bool 'IsPreMultiplied' to the constructor of Color (using telescoping constructor pattern if necessary). We can default it to the scenario most users will require, then the implementations of blend modes and such can just simply use the correct implementation of the algorithms depending on circumstance. It also gives us a really simple pattern for identifying blend mode color errors -

By default all the (en|de)coders convert to and from premultiplied color to you would end up with color going off on save. Whatever we do, it should be consistent across the board. I'm very wary of adding such a value to the constructor.

> Ghetto Easy - Provide a PreMultiplyX + PostMultiplyX and let the user decide which their color is. This just seems like an awful experience for the end user, who would think that there method by which we perform lighten wouldn't be intelligent enough to know?

Color already offers `FromNonPremultiplied`and `ToNonPremultiplied`.

> Overkill Painful - Make color itself and abstract and have two derived color classes - PreMultipiedColor + PostMulipliedColor - then provide either a static class on the base color that changes the implementation based on type (oh god no) or just provide static implementations on each derived class. I hate this option don't pick it.

I agree, let's not do this.

> What do you think is the best?

Ummmmm.... Honestly, I don't know for sure. This is what is done in `Lerp`. I'm assuming that if the color's alpha component is less that 1 then the color is premultiplied then running two different algorithms. As long as the methods is well documented then I don't see an issue with that.

    public static Color Lerp(Color source, Color destination, float amount)
    {
        amount = amount.Clamp(0f, 1f);

        if (Math.Abs(source.A - 1) < Epsilon && Math.Abs(destination.A - 1) < Epsilon)
        {
            return source + ((destination - source) * amount);
        }

        // Premultiplied.
        return (source * (1 - amount)) + destination;
    }

However we would have to carefully check the output of such methods to ensure consistency.

christopherbauer:
Isn't the A value completely independent of whether or not you're dealing with pre or post multiplied colors? That's why you can still convert to and from post once you've gone to pre (albeit lossily). Is `Lerp` accurate for this scenario?

Remember the constructor could have a default value so most people wouldn't notice, and changing implementations of converters would be fairly straightforward (based on which conversion you were doing).

If anyone, including us, plan to work with colors in any way pre or post status is something that belongs inside the color itself so that we can make intelligent and accurate decisions.

JimBobSquarePants:
As far as I understand it, my `Lerp `is correct. You can see it in action in the filter tests. `Vignette`, `Blend `and `Polaroid` all use it. I can't remember where I found the correct example but if I use either formula for both scenarios the result is incorrect.

Perhaps a readonly property? `IsPremultiplied` should be used which is toggled internally when `FromNonPremultiplied `and `ToNonPremultiplied`. That way the user can know for sure?

christopherbauer:
RE: Lerp accuracy, I think we need a scholarly article or something. At least according to http://stackoverflow.com/questions/9147973/how-do-i-know-if-a-image-is-premultiplied-alpha, you can't know if it is except in one specific case, and it doesn't seem like the Lerp impl is checking for the non ambiguous case.

What's I'm proposing is (quick switch here because bool is weird for these two scenarios):
`
public enum Multiply {
       Pre,
       Post,
}
public Color(float r, float g, float b, float a, Multiply colorMultiply=Multiply.Pre)
            : this()
        {
            this.backingVector = new Vector4(r, g, b, a);
            this.ColorMultiply = colorMultiply;
        }`

Most users would be initializing colors as pre-multiplied because most image formats use pre-multiplied, if I remember from some of the documentation I've read. If they know for a fact it will be Post they can initialize it as such. The implementations of color blending modes can perform a check for IsPreMultiplied and perform their work much easier. For  "ToNonPremultiplied" we can throw an exception ("Already non pre-multiplied!") or ignore calls to it if the color is already set to post-multiplied.

JimBobSquarePants:
The amount of raw information in the spec _should _really be enough  unfortunately I simply don't understand it enough. I think we should try to set our color struct and all methods to work in exactly the same way as the W3 rgba color.

https://www.w3.org/TR/compositing-1/

christopherbauer:
The spec is precisely why we're having this conversation - the blend mode needs to know whether the color is pre or post in order to determine how the mixing occurs in relation to backdrop alpha. I wish the spec provided more guidance for that but they seem to be assuming - somehow - you know what the pre or post status of the working color is prior to blending. We need that information somehow as well.

JimBobSquarePants:
OK then, well let's say this.... 

Any static method attached to `Color `expects premultiplied alpha. Anything higher up in the API that takes a color as a `Color `parameter will first convert using `Color.FromNonPremultiplied(color);` 

I've just realised that we already do this in 

``` csharp
public static Image BackgroundColor(this Image source, Color color, ProgressEventHandler progressHandler = null)
{
}
```

What I need to know now though is why does my premultiplied `Lerp` algorithm not work with an opaque color?

Would that be sensible?

christopherbauer:
Initialization is the time at which we would know the most about the color's origin (pre or post). It dramatically simplifies all the blending mode confusion, and it means that we can intelligently ignore someone doing something stupid like `myColor = myColor.ToPostMultiply().ToPostMultiply().ToPostMultiply().ToPostMultiply();` (if such a thing exists).

Blend modes can even perform the backdrop blending in 1 function that all blend modes call, providing just the implementation of the mixing function as the final argument.

Is there a reason this isn't a good idea?

JimBobSquarePants:
```
myColor.ToPostMultiply().ToPostMultiply().ToPostMultiply().ToPostMultiply();
```

You just know someone will do that!

Ok let's go with the enum in the constructor with a default value. Thanks for being so patient!

christopherbauer:
We both just want IP to be successful and easy to use - I thought it was a thought-provoking discussion! I'll probably do the color struct change in a separate branch and merge that, then fix the implementation in this branch.

JimBobSquarePants:
Hey @christopherbauer I got rid of premultiplied as the default color format so this should all be a lot easier now. Whoop!

Joddsson:
Hey guys, I'm trying to get into implementing a new blend mode but I'm running into some trouble. I'm fairly new to this project so I have a fairly silly question to ask :) 

Say I have implemented a new blend mode, how would I go about testing it? I guess I'm confused because it's not (obviously) a part of the API. Hope this makes sense :)

PR #330, opened 2016-02-15
Update Image.cs

MasoudRahmani:
Before reading stream for header , position should be changed to 0 if the passed stream were read before


JimBobSquarePants:
Ah good catch! Cheers! :+1:

PR #332, opened 2016-02-24
Fix measure string (framework)

khiemtong:
Updated PR from #328 

Target framework branch.


JimBobSquarePants:
Brilliant. Thanks! :+1:

PR #336, opened 2016-03-04
Made some corrections to Crop()

rold2007:
By recently using the Crop() method in PixelMode I realized that the Right/Bottom parameters were interpreted as Width/Height. So I created a unit test to prepare the correction and noticed that the PercentageMode also had some issues which I corrected too.

I'm not used to do PR so let me know if there's anything wrong in the way I did it.

### DIFF
--- a/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
+++ b/src/ImageProcessor.UnitTests/ImageProcessor.UnitTests.csproj
@@ -79,6 +79,7 @@
     <Compile Include="Imaging\ImageLayerUnitTests.cs" />
     <Compile Include="Metadata\TestBigEndianBitConverter.cs" />
     <Compile Include="Metadata\TestLittleEndianBitConverter.cs" />
+    <Compile Include="Processors\CropTests.cs" />
     <Compile Include="Processors\FormatTests.cs" />
     <Compile Include="Processors\MetaTests.cs" />
     <Compile Include="Processors\QualityTests.cs" />
--- a/src/ImageProcessor.UnitTests/Processors/CropTests.cs
+++ b/src/ImageProcessor.UnitTests/Processors/CropTests.cs
@@ -1,0 +1,100 @@
+﻿namespace ImageProcessor.UnitTests.Processors
+   {
+   using ImageProcessor.Common.Exceptions;
+   using ImageProcessor.Processors;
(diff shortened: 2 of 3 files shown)


rold2007:
I forgot to mention that the current Crop() method modifies the CropLayer sent in input (only in Percentage mode). This seems a very unusual behavior to me. I addressed this when I created the unit test but I didn't want to modify the code as I think the development has moved to the Core branch.

JimBobSquarePants:
Oh wow, thanks for this! Would it be possible for you to upload some comparison images generated using the old and new formulas? It's difficult to see what the difference will be.

rold2007:
I put up a small example, see the attachment. You'll see that I make 2 pairs of crops with equivalent values (but one in Percentage and the other in Pixels). There's a 512x512 sample image to play with. In the original code the Percentage code returns width/height of 0 hence there is no result image (causes exception). And using the Pixels mode the result isn't of the expected size (in width **and** in height), and both results should have a **different** size, which is not the case. 
With the new code we get the expected crop size of 256x128 and 256x320 for both crop mode.
Let me know if you have any question about the code (but it is quite simple really), or if you want more/different tests.
[CropExample.zip](https://github.com/JimBobSquarePants/ImageProcessor/files/159734/CropExample.zip)

JimBobSquarePants:
Just a note to say, I've not abandoned this. Just need to set aside the time to properly evaluate as it's a significant change.

JimBobSquarePants:
Hi there, 

I had a look at the code changes you made and it appears you misinterpreted the way the percentage mode works.

If you look at the documentation, when in percentage mode the right/bottom values represent the percentage to remove from those sides, not the width and height.

http://imageprocessor.org/imageprocessor/imagefactory/crop/#usage

I've just added a unit test in the following [commit](https://github.com/JimBobSquarePants/ImageProcessor/commit/045ac21153400d72058a1aaf673b71e4bb7f0bc1) that tests my logic conforms to those values and everything passes as expected.

I did. however, fix the code that changed the `CropLayer`. That was most certainly a bug so thanks for that! 

I'll be doing a bit of a tidy up of that branch soon with a new release soon.

Cheers

James

rold2007:
Oh, sorry about my mistake ! Thanks for taking the time to check my changes.

JimBobSquarePants:
No worries at all. Thanks for taking the time and effort to create a PR :+1:

Issue #324, opened 2016-02-08
Rewrite the Jpeg (De/En)coder


JimBobSquarePants:
The current Jpeg tools within the library are too slow and buggy, see #307 (Also cannot process 1x1 image)

I need to rewrite them or the library will be unusable for large images. I've been basing my work on https://github.com/briandonahue/FluxJpeg.Core but I cannot get it to work. (See jpeg branch) This is not going well...

Perhaps it would be possible to translate https://golang.org/src/image/jpeg/ from Golang, that, however is beyond my expertise.

This is a showstopper as far as I'm concerned and I really need help.

voidstar69:
@JimBobSquarePants Do you have a feel for whether it is possible to get FluxJpeg.Core working? i.e. how close is the jpeg branch to a working jpeg decoder/encoder? I cannot tell by looking at that branch - are there jpeg tests that fail?

JimBobSquarePants:
@voidstar69   It's not got any tests yet.... (I know! shocking :scream:)  as FluxJpeg doesn't have any unit tests. I've been doing an input/output comparison using a hex editor so far. I'm gonna see if I can pilfer some examples from the golang repo or libjpeg itself.

I _think_ the encoder is pretty much correct except for the actual compression part. All the marker headers before then seem ok. Working a bit blind really.

JimBobSquarePants:
@voidstar69 I've just added a unit test for the Forward DCT method and that seems to check out correctly. It's either the Huffman encoder or my looping strategy when splitting up the image into 8x8 blocks that's the issue now.

https://github.com/JimBobSquarePants/ImageProcessor/commit/7933d5fcf4da5bf8eaf42e9e0d6709a70d68ce7d

voidstar69:
@JimBobSquarePants Could you test FluxJpeg against the existing libjpeg library, writing tests to compare the Huffman encoder, Forward DCT method, and any other steps? Assuming that libjpeg produces the correct result, it might point out where the issue is in FluxJpeg.

I am afraid that I won't have time to work on this for a few days, maybe until the weekend.

JimBobSquarePants:
@voidstar69   I've been doing a physical comparison for the most part. Testing the Huffman Encoder is difficult though since it's so tightly coupled. It'd be great if you could figure this out. I'm scratching my head! :confounded:

voidstar69:
[...]

PR #345, opened 2016-03-15
gojpeg merge

mweber26:
Here is the cleaned up version of the gojpeg conversion (targeting #324)

- I formatted the code as best I can to match the existing project
- The Encoder and Decoder class have all non-essential fields/methods private
- All jpeg helper classes are marked internal to prevent external access

Let me know if you want any more changes to encapsulation or formatting.


JimBobSquarePants:
I'm good to merge this as-is. I can apply my particular OCD to the code after (I enjoy that kind of thing) It'll help me to understand properly what is going on when I add documentation anyway. 

Thanks again for this!

JimBobSquarePants:
Hi @mweber26 

As I'm going through the library to port to a generic Image type I've just noticed that there are a couple of fields in the jpeg decoder that are not used.

``` c#
    private byte[] blackPix;
    private int blackStride;
```

https://github.com/JimBobSquarePants/ImageProcessor/blob/Core/src/ImageProcessorCore/Formats/Jpg/JpegDecoderCore.cs#L322

Have you any idea what they are for? I can't figure it out from the golang source.

https://github.com/golang/go/blob/0104a31b8fbcbe52728a08867b26415d282c35d2/src/image/jpeg/reader.go#L126

I'm going to be doing a full cleanup of the class to match the styles of the other formats so I'd like to make sure everything is ok first.

mweber26:
I am pretty sure that they are used for an adobe extension to support CMKY/YPrPbK inside 3 component JPEG.

JimBobSquarePants:
Thanks for that! :+1: 

I reckon I can do without it for now then and go ahead with a cleanup. 

I'm going to be attempting to port it across to match the other encoders in my new generic image branch here. They were easy but this is a different beast. 

https://github.com/JimBobSquarePants/ImageProcessor/tree/2540005d263d7cd223413b314faafe322eee30a5/src/ImageProcessorCore/Formats

I may need to call on you if that's ok?

mweber26:
Sure, I haven't looked at the new encoder framework, but let me know if you run into any problems.

JimBobSquarePants:
Thanks. Appreciate it!

PR #346, opened 2016-03-15
Add 4:4:4 encoding to jpeg encoder.

mweber26:
Add support for 444 encoding in jpeg encoder.  Use it by default (if not set) when quality is gte 80.


JimBobSquarePants:
Excellent. Thanks!

PR #318, opened 2016-02-01
Adding basic middleware and web test projects.

PromontoryProtean:
This is a very basic ImageProcessorCore.Web project that contains middleware that does resizing of images in memory as determined by querystring parameters. It is a rough start to use as a scaffold to layer enhanced functionality onto. I also added an MVC test website that closely mirrors the test website in the current V2 repository. I did not include the file monitor code from _Layout.cshtml, however, because I wasn't sure how to replicate it in the the new aspnet core environment.

PR #348, opened 2016-03-16
Adding basic middleware and web test projects.

PromontoryProtean:
This pull request supersedes #318. I will close #318 if this works as intended.


PromontoryProtean:
Hey Jim, I was tied up with a client migration but finally had some time to revisit this. Not sure if I went about it correctly. Couldn't figure out a way to change the target branch on my original pull request, so I created a new pull request after creating a new Core-Web branch in my fork. Not sure if the old conversation will be lost. Hopefully this will work.

PromontoryProtean:
Just checking back on this and if you changed your mind. Curious what the idea you mentioned for the web branch is.

JimBobSquarePants:
Nah, not changed my mind. Just had a couple of big bugs in the jpeg decoder to get fixed first. 

My approach will actually be laughably simple. 

Something like.

``` csharp
public class Resize : IWebImageProcessor {

    public int Index GetIndex (NameValueCollection querystring){
        // Calculate the position in the querystring params.
        // Parse querystring to get width , height or other properties which are private. 
    }    
   
    public Image Process (Image image){
    
        // width, height are private 
         return image.Resize(this.width, this.height);
    }
}
```

I'll have an extension method to `Image `that loops through all known `IWebImageProcessor` types and assigns the correct one.

Something like that anyway...

JimBobSquarePants:
Think I best merge this before everything changes for RC2! Thanks so much for your patience and contribution.

PR #352, opened 2016-03-22
JPEG - Fix to read of final bytes from buffer.

mweber26:


JimBobSquarePants:
I merged this with the other PR, so I'll close this.

PR #356, opened 2016-03-27
Clarified c++ package requirements for WebP plugin

karl-sjogren:
Clarified which versions of the package require which version of the c++ package since a lot of people are still using the earlier versions and might need to redeploy it on new servers.


JimBobSquarePants:
Absolutely perfect thank you!

Issue #363, opened 2016-04-13
ImageProcessorCore .NET Framework 4.5.2 build


zabulus:
I've needed ImageProcessorCore library for using in our .NET Framework 4.5.2. And I've successfully built it. Will you accept PR for net452 target for the library?

JimBobSquarePants:
Cool! :sunglasses:  How did you get past the need for the vectors library? I didn't know it was installable on 4.5.2.

I'm assuming the need for this is so that you can use the same codebase in your application for both .NET Core and .NET 4.5.2?

If so, then yeah. I'd like to setup the configuration to support the full framework also.

zabulus:
>Cool! :sunglasses: How did you get past the need for the vectors library? I didn't know it was installable on 4.5.2.

According to this article: https://mariantines.com/2015/08/24/simd-support-in-net-framework/
>  For .NET Framework 4.5 both fixed and variable length type are in the same NuGet package of version 4.0.0.

So I simply added dependency for 4.0.0 version. With some little magic around System.Runtime it begun build. Though, I needed additional tweak in my dependant project.

> I'm assuming the need for this is so that you can use the same codebase in your application for both .NET Core and .NET 4.5.2?

Not exactly. We are using System.Drawing for saving PNG files in memory and profiler shows that it takes big amount of time. I'm looking 3-rd party libraries for performant PNG encoder or at least with tweaks of compression to decrease it.

zabulus:
Just pushed:
https://github.com/JimBobSquarePants/ImageProcessor/pull/364

razblack:
Hi, I want to say there is a problem with moving down to 4.5.2  atleast with regards to DNX.

I tried to move down to 4.5.1 for the current project I'm working on and it was failing quite badly with..

> "Could not load file or assembly 'System.IO.Compression, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'

In particular, my input file is PNG and I wanted to convert down for thumbnails to JPG... eventually to Base64String for blobs...:

```
using (FileStream stream = File.OpenRead(filePath))
using (Image image = new Image(stream))
using (FileStream output = File.OpenWrite(storagePath  + blobName + "_blob"))
{
	image.Resize(320, 180).SaveAsJpeg(output);
}
```
at using (Image image = new Image(stream))

[...]

PR #364, opened 2016-04-13
Add support for .NET 4.5.2

zabulus:


JimBobSquarePants:
@Jeavon Can you do me a favour and cast your eye over this please. You know waaaay more about the new compile/build process than me.

Jeavon:
@zabulus I would like to understand that if you are targeting .Net Framework v4.5.2 why would you choose to use ImageProcessor.Core rather than mature ImageProcessor Framework?

zabulus:
From https://github.com/JimBobSquarePants/ImageProcessor/issues/363
> Not exactly. We are using System.Drawing for saving PNG files in memory and profiler shows that it takes big amount of time. I'm looking 3-rd party libraries for performant PNG encoder or at least with tweaks of compression to decrease it.

Jeavon:
@zabulus did you get this to actually work with framework 4.5.2? I have experimented and while your change to project.json allows a NuGet package to compiled with 4.5.2 binaries in practice I'm unable to use it as System.Numerics.Vectors does not install in a 4.5.2 project, after a quick look at it I think it requires a minimum of 4.6...

Jeavon:
@zabulus I see you mentioned an "additional tweak" in #363 what was that?

zabulus:
I've manually added ```System.Runtime`` as reference. VS marked it as warning but swallowed.
As for ```System.Numerics.Vectors```. I've used nuget package v. 4.0.0 that built for 4.5.2.

JimBobSquarePants:
According to this article it should run on Net4.5.2 when Vectors are installed via Nuget. 

We probably just need to tweak the config appropriately to get System.Runtime correct. 

https://mariantines.com/2015/08/24/simd-support-in-net-framework/

JimBobSquarePants:
Just did some local testing and this all seems to work. Tanks @zabulus for the PR and thanks @Jeavon for testing.

JimBobSquarePants:
This isn't working. I cannot get the nuget package to install into a NET 4.5.2 applicatoin

KLuuKer:
I also tested it last night seems you need to manually install `System.Numerics.Vectors` version 4.0.0
and after that it SHOULD install the ImageProcessorCore package

PR #376, opened 2016-05-04
Support for separate disk cache and bug fixes

kendallb:
For the project I am looking to use this tool for, I want to cache the disk based images for a long time (a year is a good default), however I need to make sure the browser cache expires in a reasonable amount of time and attempts to re-validate the image with the source in case it has changed. The code currently sends a value of 365 days for the images to be cached in the browser, so if an image is update before it expires in the browser, the user won't ever see the updated version.

Hence I think it is better to separate the amount of time it is cached on disk, and the amount of time it is cached in the browser.

I also fixed the expiration time stamp for the disk cache, and fixed the MVC sample project.

I also fixed up compile issues for the Framework branch and recovered missing images. Some unit tests still fail but I am not sure how to fix those (quite possibly the unit test images I recovered are incorrect and the correct ones need to be checked into Git).

### DIFF
--- a/src/ImageProcessor.Web/Caching/IImageCache.cs
+++ b/src/ImageProcessor.Web/Caching/IImageCache.cs
@@ -36,6 +36,11 @@ namespace ImageProcessor.Web.Caching
         int MaxDays { get; set; }
 
         /// <summary>
+        /// Gets or sets the maximum number of days to cache the image in the browser.
+        /// </summary>
+        int BrowserMaxDays { get; set; }
+
+        /// <summary>
         /// Gets a value indicating whether the image is new or updated in an asynchronous manner.
         /// </summary>
         /// <returns>
--- a/src/ImageProcessor.Web/Caching/ImageCacheBase.cs
+++ b/src/ImageProcessor.Web/Caching/ImageCacheBase.cs
@@ -191,7 +191,7 @@ namespace ImageProcessor.Web.Caching
         /// </returns>
         protected virtual bool IsExpired(DateTime creationDate)
         {
-            return creationDate.AddDays(this.MaxDays) < DateTime.UtcNow.AddDays(-this.MaxDays);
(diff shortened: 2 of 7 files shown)


JimBobSquarePants:
There's something gone badly wrong with this branch, I think my git gui let me down somewhat.

I was able to compile everything on my machine but as a test I redownloaded and several files were different.

I'm having to make a couple of changes to the solution to ensure everything works and they're different enough to mean your changes to the solution itself will most likely conflict. This means you will have to pull the Framework branch branch again once I'm done.

I'm really sorry about this...

kendallb:
No worries. If you simply need to make changes to project files and the like I can easily rebase my changes on top. I wrote a tool for diffing and merging VS project files that I use all the time so I can easily merge any changes you make.

JimBobSquarePants:
Cheers! It looks like Nunit has jumped the shark a little too in that it doesn't support relative paths anymore so I'm gonna have to go through and fix the unit tests. I'll push the repo first with the solution fixes.

JimBobSquarePants:
That's the necessary changes pushed. The solution builds again and Images are restored. I've not implemented the changes you did in the MVC project.

kendallb:
I have rebased all my changes against your changes and it all compiles clean, so you should be able to pull my stuff over now.

JimBobSquarePants:
Merged. Thanks muchly for your help :beer:

kendallb:
Awesome, thanks for the great tool! Any idea when those changes would end up in the official NuGet version?

JimBobSquarePants:
No worries :smile: 

 I'm making pretty good progress on my milestone [issues](https://github.com/JimBobSquarePants/ImageProcessor/milestones/Framework%20Consolidation) so hopefully not too long. I might have to leave the Amazon S3 til later and do a couple of releases.

PR #381, opened 2016-05-06
Creates a cancelable event to validate the request

Shazwazza:
Creates a cancelable event to validate the request IP does the processing. This can be used to cancel the request based on parameters or to directly manipulate the parameters. This deprecates the ProcessQuerystringEventHandler and ProcessQuerystringEventHandler since events should not really return an object like string since that means multiple event handlers cannot effectively subscribe, instead the event should contain an args object that can be manipulated by multiple handlers if they are registered.


JimBobSquarePants:
Thanks for this! This looks good so I'll merge it in. I was hoping nobody had noticed I was returning the string. The consequence of late night coding!

PR #387, opened 2016-05-19
Optimized Rotate() to use pre calculated rotation matrix

icanhasjonas:


JimBobSquarePants:
@icanhasjonas Ah brilliant! It's the simple wins like this I like the best :smile:

icanhasjonas:
Well, to be honest, I just couldn't resist helping out when the Jim rocks Agent Archer as avatar.

JimBobSquarePants:
It gets me all the PR's...

Issue #392, opened 2016-05-27
bmp images with negative height produce Exception when decoding


pammann-work:
Exception: array dimensions exceeded supported range
at: BmpDecoderCore.cs
float[] imageData = new float[this.infoHeader.Width * this.infoHeader.Height * 4];

due to negative this.infoHeader.Height

pammann-work:
for negative height in bmp see here:
see http://www.drdobbs.com/architecture-and-design/the-bmp-file-format-part-1/184409517

JimBobSquarePants:
Thanks for submitting a fix. I'll close this off now :smile:

PR #393, opened 2016-05-27
#392 add support for negative height in bmp files

pammann-work:
add fix for #392


JimBobSquarePants:
Ah thanks for this. Excellent work! :+1:

Issue #395, opened 2016-05-30
bmp/gif save to stream will close the stream


pammann-work:
MemoryStream ms = new MemoryStream();
image.SaveAsBmp(ms);

will lead to a closed memory stream

and this will then not work:

using (var dest = new FileStream(destFileName, FileMode.Create, FileAccess.Write, FileShare.None, bufferSize: 16384))
{
  ms.CopyTo(dest);
}
-> Exception: "Cannot access a closed Stream"

JimBobSquarePants:
These are the kinds of thing unit tests miss and only real usage will catch. That' one of the reasons I'm running such a long alpha. Thank for this!

PR #396, opened 2016-05-30
#395 fix

pammann-work:
add a fix by not using the using(...) construct, which will keep the destination buffer (e.g. MemoryStream) open, which is essential when using MemoryStream


JimBobSquarePants:
Cheers! :smile:

Issue #408, opened 2016-06-22
ImageBase API design and unmanaged memory.


JimBobSquarePants:
This issue stems from #404 

In the current codebase the `ImageBase` class implements the `IDisposable` interface combined with  `GCHandle` to pin the pixels array in memory for the lifetime of the image. You can see that code [here](https://github.com/JimBobSquarePants/ImageProcessor/blob/16779a3ac82e4d9ca67d4b9dd318b443c12e0949/src/ImageProcessorCore/IImageBase.cs).

This has been done in order to allow accessing the individual colors using pointers to vastly increase performance (~44%) over a managed equivalent of the class. 

Now this approach causes me a little worry about the unmanaged resources for a few reasons.

1. I Keep the pixels pinned for the lifetime of the image. I don't know how that will affect performance on a server.
2. I don't know how long unmanaged references are kept in memory when you assign a new image to an existing reference. See the [example code](https://github.com/JimBobSquarePants/ImageProcessor/blob/16779a3ac82e4d9ca67d4b9dd318b443c12e0949/src/ImageProcessorCore/ImageExtensions.cs#L169) where I reassign an image without disposing.
3. I don't even know for sure whether I can get away without disposing. AFAIK there's no guarantee that a finalizer will get called.

So what are my options?

1. Go back to using a managed pixel array (Too slow - Infeasible)
2. Explicitly add a Lock/Unlock method (Developers recognise `IDisposable` and might forget to Unlock)
3. Continue as is if it is ok
4. Something else...

Any input is most welcome. This is a fundamental part of the API.

SamuelEnglard:
First let me answer your 3 questions:

1. I hate to use the line but... "It depends"
2. The unmanaged references are around till you get rid of them aka till `Dispose` is called. If you don't call it something else either has to or it will be around till the process dies.
3. There are really only two reasons a finalizer will not be called: The process exited before it could, in which case the memory is released anyways; Or another finalizer get's stuck in an infinite loop, in which case the process is in real trouble anyways.

JimBobSquarePants:
Thanks for the input. I look forward to seeing what else you can add :+1:

xoofx:
[...]

PR #409, opened 2016-06-23
Explicitly list individual NuGet dependencies

mellinoe:
This replaces the "NETStandard.Library" reference with individual references to the libraries that are actually used from the standard library. This gives people who are consuming the package the freedom to be more granular about their dependencies.


JimBobSquarePants:
That looks much more sensible than what I had thanks!

While I have your attention, could you please do me a favour and have a look at this issue #408 if you have a moment. Got a bit of a fundamental API design issue I need to fix.

PR #411, opened 2016-06-23
Skew

srekvin:
I managed to make skew work on:
image.Skew(10, 20, this.ProgressUpdate)
image.Skew(-10, -20, this.ProgressUpdate)

But it bugs on
image.Skew(-10, 20, this.ProgressUpdate)
image.Skew(10, -20, this.ProgressUpdate)

I will continue to try to fix this.

Since this is a type of skew which is on pixel level. 1 to 1. You must not lose or duplicate any pixels.

I have also remove a lot of memory used.

Dont put this into the core.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
--- a/src/ImageProcessorCore/Common/Helpers/ImageMaths.cs
+++ b/src/ImageProcessorCore/Common/Helpers/ImageMaths.cs
@@ -148,8 +148,8 @@ namespace ImageProcessorCore
             Vector2 rightBottom = Vector2.Transform(new Vector2(rectangle.Right, rectangle.Bottom), matrix);
 
             Vector2[] allCorners = { leftTop, rightTop, leftBottom, rightBottom };
-            float extentX = allCorners.Select(v => v.X).Max() - allCorners.Select(v => v.X).Min();
-            float extentY = allCorners.Select(v => v.Y).Max() - allCorners.Select(v => v.Y).Min();
+            float extentX = allCorners.Select(v => v.X).Max()- allCorners.Select(v => v.X).Min();
+            float extentY = allCorners.Select(v => v.Y).Max()- allCorners.Select(v => v.Y).Min();
             return new Rectangle(0, 0, (int)extentX, (int)extentY);
         }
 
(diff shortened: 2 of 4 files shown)


JimBobSquarePants:
Excellent. It's great to see you making progress! :+1: 

I think some of the changes you have made can be carried over to reduce the memory used on the rotate method also.

I'm going to be pushing a massive change to the project very soon which move a lot of files around and changes the way pixels are accessed to reduce overall memory pressure. You'll want to update you code from the source before continuing.

JimBobSquarePants:
Have you had a look at the transform matrices available in the System.Numerics.Vectors namespace. I'm sure they should provide all you need. I've just used them wrong.

srekvin:
Yes. I working with 3 versions. One which is not using the system.numerics.
1. This is only looking at it on the pixel level. the one in the pull request. 
2. The other method is using system.numerics.vectors.  The scale is not correct. I will look at how the new size target is created.  
3. This is use float. And works allmost as Number 1. And works bit better than nr 1. But the code is not readably.
There is one thing I dont undertstand about the code.  Why cloning the picture to firstpass? 
I will contine looking at it in the weekend.

24. jun. 2016 00.38 skrev James Jackson-South <notifications@github.com>:Have you had a look at the transform matrices available in the System.Numerics.Vectors namespace. I'm sure they should provide all you need. I've just used them wrong.

JimBobSquarePants:
Cool thanks, I'd like to get a System.Numerics version working as that gives me the SIMD performance boost.

>There is one thing I dont undertstand about the code.  Why cloning the picture to firstpass?

I cloned because the transformed points I was getting back from my calculation were outside the dimensions of the source image and led to my image being cut off. It was a sloppy workaround.

JimBobSquarePants:
Is this now a working version then? If so, awesome! I can use it to improve the memory usage in rotate. 

Don't worry about the Appveyor checks just now, I can pull, and do the necessary changes if you like?

JimBobSquarePants:
You mean the numbers I was passing to `Point.Skew`?

srekvin:
Hi. 
My code is now ready for review. It is now working. I found a bug i target resize. It gets to small. Look for invSkew variable. If rotate uses the same code the formula for target size is also worng, but it works because it is 90 deg beteen x,y axsis.
I hope this was the fault.
There are some lame if-s i the code which can be changed.
26. jun. 2016 23.13 skrev James Jackson-South <notifications@github.com>:Is this now a working version then? If so, awesome! I can use it to improve the memory usage in rotate. 

Don't worry about the Appveyor checks just now, I can pull, and do the necessary changes if you like?

JimBobSquarePants:
Hey @srekvin I've just merged and tested your Skew code and it works great! Thanks! :+1: 

Could you do me a huge favor and have a look at RotateProcessor.cs to see where I have gone wrong there. I've tried to port your approach across but the rotated image is off-center. I'm sure it's something simple I've missed.

srekvin:
I can look at it. 
28. jun. 2016 08.49 skrev James Jackson-South <notifications@github.com>:Hey @srekvin I've just merged and tested your Skew code and it works great! Thanks! 👍 

Could you do me a huge favor and have a look at RotateProcessor.cs to see where I have gone wrong there. I've tried to port your approach across but the rotated image is off-center. I'm sure it's something simple I've missed.

PR #421, opened 2016-07-01
Skew and rotate

srekvin:
hi,

I made rotate and skew to use almost the same code. 

In future, a function like this

var m  =  Matrix3x2.CreateSkew(1.2f,-0,2f)
 image.ApplyMatrix(m, this.ProgressUpdate) .Save(output);

Then you can create a matirx which both skew,rotate,scale and translates in one operation..

have a nice day.


JimBobSquarePants:
Brilliant stuff. I moved the helper methods into an abstract class and it all works well. Cheers! :beers: 

Oh just to say, I'd update your fork from my latest code. There's been a few changes that cause conflicts when merging.

Issue #428, opened 2016-07-13
AzureBlobCache.streamCachedImage disables browser caching


JasonElkin:
Creating a new response and streaming the blob directly ([here](https://github.com/JimBobSquarePants/ImageProcessor/blob/Framework/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs#L387)) breaks the cache/ETag headers and ability to send a 304 response.

Happy to have a stab at fixing this - should just be a case of mapping the headers between the requests.

JimBobSquarePants:
Hi Jason,

Thanks for this.

Yeah please do. The functionality is something that was requested but that I have personally never used so I wont be able to test properly.

JimBobSquarePants:
This is now merged but I'll keep it open until I ship again. Thanks for your efforts!

JimBobSquarePants:
@JasonElkin That's the code released now. Thanks again for your efforts. Much appreciated! :+1:

PR #431, opened 2016-07-15
304 passthrough for AzureBlobCache.streamCachedImage

JasonElkin:
Fix for #428 

Maps HTTP headers between browser and Azure CDN to enable automatic handling of 304s

### DIFF
--- a/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
+++ b/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
@@ -231,7 +231,7 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
             await blockBlob.UploadFromStreamAsync(stream);
 
             blockBlob.Properties.ContentType = contentType;
-            blockBlob.Properties.CacheControl = string.Format("public, max-age={0}", this.MaxDays * 86400);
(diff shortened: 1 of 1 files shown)


JasonElkin:
While testing I noticed that the AzureBlobCache makes a lot of unnecessary (and taxing) calls to the Azure Blob Storage API.

I was going to raise it as a separate issue but I accidentally added the fix to this Pull Request.

Anyways, the changes save ≈500ms (at least in my test environment).

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -378,24 +392,58 @@ public override void RewritePath(HttpContext context)
...
+                    {
+                        HttpWebResponse response;
+                        try
                         {
-                            using (HttpWebResponse response = (HttpWebResponse)request.GetResponse())
+                            response = (HttpWebResponse)request.GetResponse();
+                        }
+                        catch (WebException ex)
+                        {
+                            // A 304 is not an error
```
JimBobSquarePants: Can't believe I missed that. I need to check elsewhere in my codebase now for similar!

JimBobSquarePants:
Great stuff! Thanks for this :+1: The added performance benefits are massive also so very pleased with the work. 

Very much appreciated!

alindgren:
@JasonElkin awesome! I noticed the calls too - this should help us out a lot!

Issue #432, opened 2016-07-15
[Proposal] A non-destructive way to support AzureBlobCache settings customisation


Philo:
We've previously discussed (#369) the ability to support customisation of the settings values for the ```ImageCacheBase``` and specifically the ```AzureBlobCache```.  

At present all configuration logic happens within the constructor, whether the Settings dictionary to read and applied to internal properties, this method prevents any derived implementations from altering those base settings before they are applied.  

From our previous discussion you were not inclined to make any significant changes the ImageProcessor as you were starting on Core support, I can totally understand and accept that, I do however have a proposal for a minor change that will allow for the ```AzureBlobCache``` and ```ImageCacheBase``` to support a customisation path for the scenario where you want to store configuration details elsewhere.

My proposal is to introduce a means to augment the ```ImageCacheBase``` with two new methods:

```c#
private void AugmentSettingsCore(Dictionary<string, string> settings)
{
  AugmentSettings(settings);
}

protected virtual void AugmentSettings(Dictionary<string, string> settings)
{
}
```
Then call ```AugmentSettingsCore(Settings)``` from the ```ImageCacheBase``` constructor:

```c#
protected ImageCacheBase(string requestPath, string fullPath, string querystring)
{
  this.RequestPath = requestPath;
  this.FullPath = fullPath;
  this.Querystring = querystring;
  this.Settings = ImageProcessorConfiguration.Instance.ImageCacheSettings;
  this.MaxDays = ImageProcessorConfiguration.Instance.ImageCacheMaxDays;
  this.BrowserMaxDays = ImageProcessorConfiguration.Instance.BrowserCacheMaxDays;
  AugmentSettingsCore(Settings);
}
```

This will allow a derived instance to act on the settings before the immediate classes (like ```AzureBlobCache```) and then allow the Settings values to be updated via custom logic.

Here is my working example (that has required me to re-implement the ```AzureBlobCache``` locally):

```xml
<appSettings>
  <add key="AzureBlobCache.CachedStorageAccount" value="<storage connection string>" />
  <add key="AzureBlobCache.CachedCDNRoot" value="https://somewhere.com" />
  <add key="AzureBlobCache.SourceStorageAccount" value="<storage connection string>" />
</appSettings>

<caching currentCache="AzureBlobCache">
  <caches>
    <cache name="AzureBlobCache" type="Test.AzureBlobCacheFromAppSettings, Test" maxDays="365" />
  </caches>
</caching>
```

[...]

PR #435, opened 2016-07-17
Customisation of Settings within ImageCacheBase

Philo:
re #432 

As suggested, I have however added the ```AugmentSettings``` method to the base ```ImageCacheBase``` so that the S3 cache can use the same method to create the cache with custom settings resolver.

### DIFF
--- a/src/ImageProcessor.Web/Caching/ImageCacheBase.cs
+++ b/src/ImageProcessor.Web/Caching/ImageCacheBase.cs
@@ -61,10 +61,23 @@ namespace ImageProcessor.Web.Caching
             this.RequestPath = requestPath;
             this.FullPath = fullPath;
             this.Querystring = querystring;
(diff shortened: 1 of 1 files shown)


Philo:
Hi @JimBobSquarePants, just curious if you have had a chance to review this pull request yet?

JimBobSquarePants:
@Philo Not yet I'm afraid. Juggling a lot of work... I will get round to it as soon as I can though and as I said before I like the approach you have taken so it should be an easy PR.

JimBobSquarePants:
@Philo Merged in now. Thanks again! :+1:

Philo:
That's great, when will this update be pushed to nuget? (if that hasn't already happened)

JimBobSquarePants:
I'm performing final check to ensure I haven't broken anything and am also looking through other open tickets. I might get to release it tonight if all goes well.

Philo:
Fantastic, thanks

PR #448, opened 2016-08-09
Changes to ImageProperty

dlemstra:
This pull request includes the following changes for ImageProperty :

It is no longer a struct.
Guard that the name of a property can not be null or empty.
Make sure that Equals and GetHashCode can handle a null value.
Added unit tests.


JimBobSquarePants:
Excellent stuff. All looks good :+1:

PR #449, opened 2016-08-12
Change encoder to use Image instead of ImageBase.

dlemstra:
Some of the encoders do a cast from ImageBase to Image to make sure they can access some properties. This patch changes the IImageEncoder to use Image instead of ImageBase and removes those casts.


JimBobSquarePants:
Ah good. I was looking to do this myself. I should probably do the same with the decoders also.

dlemstra:
You already fixed it for the decoder :)

JimBobSquarePants:
Winning! Let's merge this then!

PR #450, opened 2016-08-13
Exif support

dlemstra:
These patches will add support for reading and writing exif profiles to the JPEG encoder and decoder. The ExifProfile class can also be used by other coders in the future.


JimBobSquarePants:
This is such a massive bonus to the library. Very happy! 💯

PR #466, opened 2016-09-05
Do a null check on headers before copying them

mortenbock:
A null reference is thrown if the original response did not have the headers.


JimBobSquarePants:
Will get onto this asap. Not had my evenings this week to work on the library.

PR #481, opened 2016-09-27
Add an target of ImageProcessorCore for net45 directly

yyjdelete:
Make it allow to install the package on net45+ with much less dependencies


JimBobSquarePants:
Thanks @yyjdelete I'll have a look.

@mellinoe Could you perhaps comment on this as you submitted the original PR with the current project.json

mellinoe:
There's nothing really wrong with adding this configuration (it should build, reference, run, etc. fine), but I don't know if it's necessarily that valuable. Usually you would only want a separate build configuration if there was something different about that platform, like different code, different compilation constants, etc.  I understand that the desire is to reduce the visible dependencies, but it might also be good to understand that the library isn't really bringing in much "real code" from the dependencies. On `net45`, most of the dependencies listed will just turn into facade assemblies that type-forward to mscorlib.dll, System.dll, etc. and be otherwise harmless The tradeoff here is adding a bit of build and packaging complexity in order to avoid having those dependencies/references show up in a `net45` project.

From a higher-level perspective, I would say that these dependencies aren't really anything to worry about or try to eliminate.

JimBobSquarePants:
Soooo.... Really sorry about this but I've managed to break any open PR's in an attempt to reduce the repository size. I'm going to have to close this PR. 

That said, I don't think I will make the changes just now for the reasons Eric gave. I'll be updating everything once simplified csproj files are released anyway.

Issue #462, opened 2016-08-23
Multidimensional vs jagged arrays


omariom:
Access to jagged array elements is currently faster:

Output:
```
jagged (sec)
0.000683631317134085 

multidimensional (sec)
0.00203599641664631 
```

Microbecnhmark:
```C#
float _f;

void Main()
{
    float[][] jagged = new float[1000][];
    
    for (int i = 0; i < jagged.Length; i++)
    {
        jagged[i] = new float[1000];
    }
  	
	Measure.Time(()=>
	{
        float f = 0;

        for (int y = 0; y < 1000; y++)
        {
            var ys = jagged[y];
            for (int x = 0; x < 1000; x++)
            {
                //f = jagged[y][x]; 
                f = ys[x];
            }
        }
		_f = f;
	}).Dump("jagged (sec)");
    
    float[,] multiDim = new float[1000, 1000];

    Measure.Time(() =>
    {
        float f = 0;

        for (int y = 0; y < 1000; y++)
        for (int x = 0; x < 1000; x++)
        {
            f = multiDim[y, x];
        }

        _f = f;
    }).Dump("multidimensional (sec)");
}
```

JimBobSquarePants:
I'm trying to think where I use a multidimensional array? I've been using a 1D array throughout and mathing my way to the correct part. As far as I'm aware that should be faster than both.

omariom:
@JimBobSquarePants 

I see `float[,]` in many places. 
For example, [Convolution2PassFilter](https://github.com/JimBobSquarePants/ImageProcessor/blob/Core/src/ImageProcessorCore/Filters/Processors/Convolution/Convolution2PassFilter.cs#L23).

JimBobSquarePants:
Ah yeah, of course!

Managed to break all my edge detection code in the last few commits so I'll look into switching out as I figure out whats wrong.

JimBobSquarePants:
@omariom I've finally gotten round to switching all the code round. Thanks for the suggestion! 😄

PR #501, opened 2016-10-21
Use jagged instead of multidimensional array

dlemstra:
Use jagged instead of multidimensional array as suggested in #462.
Only create the kernels once in the filters.

### DIFF
--- a/src/ImageProcessorCore/Samplers/Processors/Convolution/Convolution2DFilter.cs
+++ b/src/ImageProcessorCore/Samplers/Processors/Convolution/Convolution2DFilter.cs
@@ -23,7 +23,7 @@ namespace ImageProcessorCore.Processors
         /// </summary>
         /// <param name="kernelX">The horizontal gradient operator.</param>
         /// <param name="kernelY">The vertical gradient operator.</param>
-        public Convolution2DFilter(float[,] kernelX, float[,] kernelY)
+        public Convolution2DFilter(float[][] kernelX, float[][] kernelY)
         {
             this.KernelX = kernelX;
             this.KernelY = kernelY;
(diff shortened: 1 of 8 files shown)


dlemstra:
I added the benchmarks and I am getting the following results on my machine (2 runs):

_before:_
```
|      Median |    StdDev | Gen 0 | Gen 1 |  Gen 2 | Bytes Allocated/Op |
|------------ |---------- |------ |------ |------- |------------------- |
| 426.0806 ms | 4.7395 ms |     - |     - | 450.00 |       8.008.666,72 |
| 427.2504 ms | 4.7152 ms |     - |     - | 496.00 |       8.338.740,79 |
```

_after:_
```
|      Median |    StdDev | Gen 0 | Gen 1 |  Gen 2 | Bytes Allocated/Op |
|------------ |---------- |------ |------ |------- |------------------- |
| 411.7357 ms | 3.7951 ms |     - |     - | 446.00 |       8.012.231,19 |
| 412.0662 ms | 4.3176 ms |     - |     - | 449.00 |       7.873.010,86 |
```
So this is only 3.6% faster but I think we should go ahead and replace the other multidimensional arrays with jagged arrays. I can do that after this one has been merged.

Can you clarify what you mean by double checking the properties and where am I not using C#6?

JimBobSquarePants:
>Can you clarify what you mean by double checking the properties and where am I not using C#6?

Yeah, instead of using `private static` etc you should be able to do something like 

`public override float[][] KernelX { get; } = { */..array here..*/}`

dlemstra:
I changed it to this so we would not create a new array each time. The C#6 syntax would create a new array each time but the array is never modified.

JimBobSquarePants:
`=>` would create a new array. I'm fairly certain simply using `=` doesn't though.

dlemstra:
It won't for this instance but it will for every new instance that is created. I think we can just share the same array between all instances and threads.

JimBobSquarePants:
Cool... I'm happy to merge this in then.

Issue #506, opened 2016-11-01
Url encoding in remote image service


ZNS:
Hello again,

This issue is related to #487 There are issues with URL decoding in the remote image service. Here's an example of passing a valid url to remote.axd **without** encoding it first.
`/remote.axd?https://external.xx.fbcdn.net/safe_image.php?d=AQBKGZEQAA7QoMzM&url=http%3A%2F%2Fwww.gotakanal.se%2Fmedia%2F1566%2Fbildspel_1440x780.jpg%3Fanchor%3Dcenter%26mode%3Dcrop%26width%3D1024%26rnd%3D130921477360000000`

This results in the following URL being parsed by the service:
`https://external.xx.fbcdn.net/safe_image.php?d=AQBKGZEQAA7QoMzM&url=http://www.gotakanal.se/media/1566/bildspel_1440x780.jpg?anchor=center&mode=crop&width=1024&rnd=130921477360000000`
Which is **incorrect**. The encoded querystring of the valid url has now been decoded.

Now you'd think the solution to this would be to first encode the url before passing it to remote.axd
`https%3A%2%2Fexternal.xx.fbcdn.net%2Fsafe_image.php%3Fd%3DAQBKGZEQAA7QoMzM%26url%3Dhttp%253A%252F%252Fwww.gotakanal.se%252Fmedia%252F1566%252Fbildspel_1440x780.jpg%253Fanchor%253Dcenter%2526mode%253Dcrop%2526width%253D1024%2526rnd%253D130921477360000000`
However, this gives the exact same incorrect result as above.

I'm not sure how to solve this. I think the best solution would be to require that the passed url is encoded. Even though this might be a breaking change, it wouldn't really make a difference since the URL is already decoded (incorrectly) today.

JimBobSquarePants:
>This results in the following URL being parsed by the service:
https://external.xx.fbcdn.net/safe_image.php?d=AQBKGZEQAA7QoMzM&url=http://www.gotakanal.se/media/1566/bildspel_1440x780.jpg?anchor=center&mode=crop&width=1024&rnd=130921477360000000
Which is incorrect. The encoded querystring of the valid url has now been decoded.

I don't follow.. Which parts of the url are incorrectly decoded?

[...]

Issue #487, opened 2016-10-05
Facebook cdn input stream not supported?


ZNS:
Hi,

I'm having an issue when trying to manipulate a facebook image using imageprocessor web I get an "Input stream not a supported format" exception. Example url:

/remote.axd?https://external.xx.fbcdn.net/safe_image.php?d=AQDA9W98lWDqbw_J&url=https%3A%2F%2Fwww.facebook.com%2Fads%2Fimage%2F%3Fd%3DAQIUo-bMpMJ8HpStbTYWpoZ4VV5R8vuGRdKtUTwLnf5fEB84xPzsPsqyIAzQF3XNpGgKaMeg1fY1lg-oYGWcPNjX56ALRncQ95FDLDVijKu4LdMYROn_o1XfYU3SFY2Bv__Nt_eFGn8nBwNCnQUruSNj?width=640&height=640&mode=crop

I get the same error with all facebook cdn images I've tried. I have added domains to whitelist and it works with other remote images. Using imageprocessor.web version 4.6.4

JimBobSquarePants:
Upgrade to the newly released 4.6.5, that correctly handles encoded urls and I've tested it against examples I've been given.

https://www.nuget.org/packages/ImageProcessor.Web/4.6.5

Let me know if you still have issues.

ZNS:
Thanks. I updated but still get the same error. Copy/Pasted the example path above into the browser. Do I need to urlencode some part of the path?

JimBobSquarePants:
Please accept my apologies. I thought I'd managed to cover all potential urls with the fix for #478 but I was mistaken. 

I've just pushed ImageProcessor .Web version 4.6.6 which I have successfully tested against your url.

This uses the value 

``` xml
<add url="external.xx.fbcdn.net/"/>
```

to security.config

Thank you for your perseverance, I really appreciate it.

ZNS:
No problem, thank you for the quick fix and also for a great job with ImageProcessor.

ZNS:
Now the URL in my first example works fine, but I'm very sorry to say there are more that don't work.. same error. A few examples:

/remote.axd?https://external.xx.fbcdn.net/safe_image.php?d=AQBgV2Q_yDUlNQKs&url=http%3A%2F%2Fresources.mynewsdesk.com%2Fimage%2Fupload%2Ft_open_graph_image%2Fdt1us8xdnvg0c89cvw7x.jpg?width=640&height=640&mode=crop

[...]

Issue #478, opened 2016-09-22
Encoding Errors on incoming image URLs


CVertex:
Hey,

I'm using ImageProcessor to process some images for our customers. Recently, people started sharing these URLs on facebook and IP is not decoding parameters correctly.

Instead of the URL https://muru.info/remote.axd?https://i.scdn.co/image/8851b36688b767d01a00c7d1f0981997db0062d2?width=512&startcolor=aad774&endcolor=ac2833

Facebook is asking for 

https://muru.info/remote.axd?https%3A%2F%2Fi.scdn.co%2Fimage%2F8851b36688b767d01a00c7d1f0981997db0062d2%3Fwidth=512&startcolor=aad774&endcolor=ac2833

and as a results, we're getting errors like so:

![image](https://cloud.githubusercontent.com/assets/120773/18739312/f73c508c-80e5-11e6-87ac-42fedfa8e436.png)

How can I fix this? Shouldn't IP automatically detect these encoded URLs?

Any help is greatly appreciated.

Thanks

JimBobSquarePants:
Hi Vijay,

This will be the line in question. Should be easy enough to test and fix.

https://github.com/JimBobSquarePants/ImageProcessor/blob/1daf15cb83cd7b715192260acfdf76aca8dc3b9f/src/ImageProcessor.Web/Services/RemoteImageService.cs#L78

I'll have a look asap, though feel free to have a peek yourself.

Cheers

James

JimBobSquarePants:
Looks like I've fixed it in source. I tested it in the test website with your encoded url.

 Could you have a look to be sure though.

JimBobSquarePants:
Fixed in latest release 4.6.5

CVertex:
@JimBobSquarePants 
I only just came back to look at this properly!
Thanks for fixing this! You're a legend

JimBobSquarePants:
No problemo 😎

PR #509, opened 2016-11-02
Urlencoding

ZNS:
See issue #506


src/ImageProcessor.Web/Helpers/UrlDecoder.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// <summary>
+    /// Helper class for decoding passed urls
+    /// </summary>
+    public class UrlDecoder
+    {
+        /// <summary>
+        /// A new instance of the <see cref="UrlDecoder"/> class.
+        /// with lazy initialization.
+        /// </summary>
+        private static readonly Lazy<UrlDecoder> Lazy = new Lazy<UrlDecoder>(() => new UrlDecoder());
```
JimBobSquarePants: This doesn't need to have the overhead of Lazy<T> a simple static class would suffice since the method is threadsafe.

tests/ImageProcessor.Web.UnitTests/UrlParserUnitTests.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    using NUnit.Framework;
+
+    /// <summary>
+    /// Url decoding tests
+    /// </summary>
+    [TestFixture]
+    public class UrlParserUnitTests
+    {
+        [Test]
```
JimBobSquarePants: Love that you have separated out the method and made it testable. I'm assuming all the urls mentioned in #506 and #487 also pass?

ZNS: It will handle all those urls yes.

src/ImageProcessor.Web/Helpers/UrlDecoder.cs
ZNS: Yeah, some lazy copy/paste there from me, pun. I'll fix that.

ZNS:
The main issue here is handling #478 I think the workaround is ok but not fool proof. What I'd like to do is to actually add a helper for url parsing which takes the raw url and returns the media path and the imageprocessor query. Then write tests for those. But I'm not sure I have the time to do that right now.

JimBobSquarePants:
I will try and have a look if I can. Won't be quick though as I'm juggling 3 separate libraries just now.

ZNS:
Found some time so I refactored it a bit. The helper class is now responsible for parsing the requestpath and imageprocessor querystring from the rawurl. This means I've written tests for parsing all kind for URL:s even for #478

JimBobSquarePants:
This is fantastic thank you! I'll pull down your changes and run some tests asap. It all looks great so far.

JimBobSquarePants:
You rock! 🤘

PR #510, opened 2016-11-07
Add optional anchor point to resize

sWW26:
Added ability to manually define the specific anchor point when resizing in `BoxPad` mode

### DIFF
--- a/src/ImageProcessor/Imaging/Resizer.cs
+++ b/src/ImageProcessor/Imaging/Resizer.cs
@@ -83,8 +83,9 @@ namespace ImageProcessor.Imaging
             int maxWidth = this.ResizeLayer.MaxSize?.Width ?? int.MaxValue;
             int maxHeight = this.ResizeLayer.MaxSize?.Height ?? int.MaxValue;
             List<Size> restrictedSizes = this.ResizeLayer.RestrictedSizes;
+            Point? anchorPoint = this.ResizeLayer.AnchorPoint;
 
-            return this.ResizeImage(source, width, height, maxWidth, maxHeight, restrictedSizes, mode, anchor, upscale, centerCoordinates, linear);
+            return this.ResizeImage(source, width, height, maxWidth, maxHeight, restrictedSizes, mode, anchor, upscale, centerCoordinates, linear, anchorPoint);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/ImageProcessor/Imaging/Resizer.cs
```diff
@@ -234,44 +238,60 @@ protected virtual Bitmap ResizeLinear(Image source, int width, int height, Recta
...
+                                case AnchorPosition.Bottom:
+                                    destinationY = height - sourceHeight;
+                                    destinationX = (width - sourceWidth) / 2;
+                                    break;
+                                case AnchorPosition.BottomLeft:
+                                    destinationY = height - sourceHeight;
+                                    destinationX = 0;
+                                    break;
+                                default:
+                                    if (centerCoordinates != null && centerCoordinates.Any())
```
dlemstra: Can you clarify why you also added this?

sWW26: It's been a long day.  Should have been removed.

src/ImageProcessor/Imaging/Resizer.cs
```diff
@@ -234,44 +239,52 @@ protected virtual Bitmap ResizeLinear(Image source, int width, int height, Recta
...
-                                break;
-                            case AnchorPosition.BottomLeft:
-                                destinationY = height - sourceHeight;
-                                destinationX = 0;
-                                break;
-                            default:
-                                destinationY = (height - sourceHeight) / 2;
-                                destinationX = (width - sourceWidth) / 2;
-                                break;
+                            destinationY = anchorPoint.Value.Y;
```
JimBobSquarePants: There's no sanitation here for the X and Y values. We'll need that before we can continue.

JimBobSquarePants:
Hi @sWW26 

Thanks for this. What's the use case? We normally have a bit of a chat on Gitter before adding features.

sWW26:
Hi,

This might not be the best way of doing this but I couldn't seem to find another way.  I needed a way to increase the size of an image by creating the new increased size container then placing the image inside it.

For example taking this image

hn

and placing it inside a larger box in a specific position

hn-new

JimBobSquarePants:
@sWW26 Doesn't look like you need to rebase anything here so will do a final check and merge over the weekend. Thanks! 💯

PR #512, opened 2016-11-08
Update configuration.md

leekelleher:
Removed duplicate `enabled="true"` attribute for the Crop plugin.


JimBobSquarePants:
Find/replace failure on my part there. Thanks!

leekelleher:
No worries, happy to help (with what little I can).
I only noticed because I'd copy-n-pasted from the docs and VS cried at me. 😺

Issue #369, opened 2016-04-22
Allow configuration values to be evaluated from appSettings and/or ConnectionStrings


Philo:
This issue is related to my attempts to utilise ImageProcessor (and the Umbraco Filesystem provider) in a project where we do not want to store any sensitive configuration information within the code base.  As we deploy our solutions to Azure we use the appSettings and connectionStrings section configurations from the Azure portal.  However image processors configuration is read from custom configuration sections that can not be managed via the Azure portal.

My proposal is to define a method by which a custom configuration object can be defined or injected into the various configured objects, either:

1. Refactor the configuration setup out of the various constructors and allow things like the cache and processor to be inherited by custom implementations that can provide the required configuration by other means (i.e. appsettings, or other custom mechanism)

2. Support a formatting within the existing configuration sections and elements that allows one to indicate that a specific configuration value should be read from an alternate source:

```xml
    <cache name="AzureBlobCache" type="ImageProcessor.Web.Plugins.AzureBlobCache.AzureBlobCache, ImageProcessor.Web.Plugins.AzureBlobCache" maxDays="365">
      <settings>
        <setting key="CachedCDNRoot" value="+connectionStrings:cachedCDNRoot" />
      </settings>
```

```xml
      <connectionStrings>
            <add name="cachedCDNRoot" connectionString="" />
      </connectionStrings>
```

Happy to open discussions on whether this is feasible and what your expectations on such an implementation would be.

JimBobSquarePants:
Hi Philo,

This sounds like a lot of work with a strong possibility of breaking backwards compatibility. My focus is now very firmly on the new Core ImageProcessor so I'm loath to add new functionality to the Framework branch as managing these changes becomes an extra burden.

Although I wrote the Umbraco FileSystemProvider I consider them to be entirely separate entities and I wrote that code in accordance to the way configuration works within Umbraco. How would you propose to manage the changes while ensuring that developers can continue with business a usual.

I must ask... Why are your source control systems not secure enough?

Philo:
I am appreciate any implementation would not be trivial and would need to be carefully considered to maintain compatibility.  

Will the new Core version support a more abstract configuration sub-system?

> I must ask... Why are your source control systems not secure enough?

[...]

PR #515, opened 2016-11-15
#369 allow the user override settings in web.config app settings

joebyrne:
using the defined appsetting format the user can override the settings, useful for deployments to Azure etc

e.g. to override the storage hostname the following app setting key would be used

`<add key="ImageProcessor.CloudImageService.Host" value="myimageblobstorage.blob.core.windows.net">`


JimBobSquarePants:
This looks like an interesting way to solve the problem. I wonder if we can (and should) expand upon it to allow overriding other configuration values?

Also, I personally use config transforms during my build process to change settings, is that something you have attempted or considered?

octaviobffernandes:
@JimBobSquarePants I use config transforms in my project to customize these settings per environment/tenant. However, it would be also good to have an enhancement like this to allow the settings to be defined on AppSettings and thus deployed using Azure templating.
Similarly to what was done here for the caching configuration https://github.com/JimBobSquarePants/ImageProcessor/pull/435

CasperTDK:
@JimBobSquarePants config transforms wont work for azure slot swapping - we can only set slot specific settings in appsettings

JimBobSquarePants:
This will need expanded upon and completed then.

joebyrne:
I'll have a look today at what is involved in implementing the other settings.

JimBobSquarePants:
@joebyrne Great, thanks! 👍

joebyrne:
I have pushed changes to allow settings from cache, security or processing config files be managed from web.config / deployment slot app settings. A future update might also be to allow other settings such as whitelists be configured also.

JimBobSquarePants:
Thanks @joebyrne I'll have a good look asap at your changes.

JimBobSquarePants:
@joebyrne Please ignore the bot; I will get round to this eventually!

JimBobSquarePants:
Thanks @joebyrne I'll get a release out asap!

merijng:
@joebyrne @JimBobSquarePants any thoughts about when this feature is getting merged to Master + NuGet? I think this is a feature that many people would like to see.

JimBobSquarePants:
It’s been released for years.

PR #525, opened 2016-12-20
Use nuget version placeholders, adds prerelease support, bumps versio…

Shazwazza:
…ns of core and web projects based on xml


JimBobSquarePants:
Aw Nice! That's much better!

Betas are for wimps though 😛

PR #526, opened 2016-12-20
If-Modified-Since fix

ismailmayat:
If date is utc as it is when coming from googlebot then we get parse
errors so added method to work around this

### DIFF
--- a/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
+++ b/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
@@ -405,7 +405,7 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
                 // Map headers to enable 304s to pass through
                 if (context.Request.Headers["If-Modified-Since"] != null)
                 {
-                    request.IfModifiedSince = DateTime.Parse(context.Request.Headers["If-Modified-Since"]);
+                    TrySetIfModifiedSinceDate(context, request);
                 }
 
                 string[] mapRequestHeaders = { "Cache-Control", "If-None-Match" };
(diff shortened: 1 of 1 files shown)


JasonElkin:
@ismailmayat do you have an example of the header as sent by Googlebot? Tried to fix this but couldn't find an example.

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -519,6 +519,42 @@ public override void RewritePath(HttpContext context)
...
+                    ifModifiedFromRequest = ifModifiedFromRequest.ToLower().Replace("utc", string.Empty);
+
+                    if (DateTime.TryParse(ifModifiedFromRequest, out ifModifiedDate))
+                    {
+                        request.IfModifiedSince = ifModifiedDate;
+                    }
+                }
+                else
+                {
+                    ImageProcessorBootstrapper.Instance.Logger.Log<string>("Unable to parse date " +
```
JimBobSquarePants: The type param here should be the parent class. Also might be an idea to add information about the request url

JimBobSquarePants:
@ismailmayat Thanks for this. Just noted a change to the logging I would make. 

@JasonElkin I'd like to see that header also. Keen to see what triggers the different response.

ismailmayat:
Guys,

So example one 

HTTP_USER_AGENT
Googlebot-Image/1.0

HTTP_IF_MODIFIED_SINCE
Sat, 17 Dec 2016 18:08:18 UTC

Taken from elmah.io

Regards

Ismail

JasonElkin:
Thanks @ismailmayat that's really interesting - the [HTTP specification ](https://tools.ietf.org/html/rfc7231#section-7.1.1.1)says that it should be "GMT" or nothing at all.

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
ismailmayat: Done

JimBobSquarePants:
GMT of GTFO 😈 

Sorry... Couldn't help myself.

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -548,8 +548,7 @@ private static void TrySetIfModifiedSinceDate(HttpContext context, HttpWebReques
                 }
                 else
                 {
-                    ImageProcessorBootstrapper.Instance.Logger.Log<string>("Unable to parse date " +
-                                                                      context.Request.Headers["If-Modified-Since"]);
+                    ImageProcessorBootstrapper.Instance.Logger.Log<string>("Unable to parse date " + context.Request.Headers["If-Modified-Since"] + " for " + context.Request.Url, "ImageProcessor.Web.Plugins.AzureBlobCache.AzureBlobCache");
```
JimBobSquarePants: So close! The logger works just like the Umbraco one.

```
ImageProcessorBootstrapper.Instance.Logger.Log<AzureBlobCache>("Unable to parse date " + context.Request.Headers["If-Modified-Since"] + " for " + context.Request.Url);
```

ismailmayat: argghh u code nazi lol, i have updated.

JimBobSquarePants:
Thanks! 💯

Issue #529, opened 2017-01-04
ContentType header Bug with storing images outside of web root


Shazwazza:
Found an issue with my PR with images outside of root :/ It doesn't set the Content Type header correctly and defaults to text/html. I guess RewritePath takes care of that automatically which uses IIS's internal native processing to do that.

In the `DiskCache.AddImageToCacheAsync` it does accept a mimetype, however that is not persisted anywhere to be re-used. The only way to take advantage of that would be to write out a json file or something in addition to the persisted image file (maybe you have a better idea). 

The easier hackier way would be to just manually map known content types based on the file extension. This might be the only option currently and is actually what the RewritePath would be doing anyways.

I'll create a PR for that but let me know what you think.

Shazwazza:
PR: https://github.com/JimBobSquarePants/ImageProcessor/pull/530

JimBobSquarePants:
Had a look at the PR. That's the sanest, most lightweight solution we have with the API as it is. I'll get a release out asap.

Shazwazza:
Will close this since the PR was pulled in

PR #530, opened 2017-01-04
ContentType header Bug with storing images outside of web root #529

Shazwazza:


JimBobSquarePants:
Fixed and released 👍 

https://www.nuget.org/packages/ImageProcessor.Web/4.7.1

PR #532, opened 2017-01-07
need to dispose Bitmap object after we are done with it.

avs099:


JimBobSquarePants:
Great catch! Thanks for finding this.

PR #541, opened 2017-01-14
Feature/better cache

JimBobSquarePants:
These changes vastly improve the caching mechanisms performance and testability. They also give us the ability to configure the caching mechanisms at a much greater level than ever before.

**Improvements**
- All caches now use a simple SHA1 hash of the request url via `CachedImageHelper`. File versioning is still supported but will now replace the existing cached file rather than create a new one. This reduces cache size. 
- Caches only perform trimming when a new image is cached. This was originally the case but somehow it got changed to per-request. (I don't know when)
- Caches can now be optionally configured to change the nested folder depth depending on requirements. This allows us to tweak performance of the trimming operation. (default to 6 for 12.92 million files)
- Trimming the cache is now optional (defaulting to true) 
- Only one trimming operation can run at a time. 

The only downside to all this is that the original cache from any prior installations is now invalid. I'll have to make this release **v4.8.0** in order to justify the changes.

I was originally going to operate on a one-in-one-out cleanup operation on the caches for performance (I still might) but that could lead to orphaned files. 

@dlemstra @nul800sebastiaan @Shazwazza @ZNS If you have the time I'd really appreciate your input on the changes. Especially on the one-in one-out approach. 

Cheers

James

### DIFF
--- a/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
+++ b/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
@@ -62,9 +62,9 @@ namespace ImageProcessor.Web.Caching
             // That name can also be used as a key for the cached image and we should be able to use
             // The characters of that hash as sub-folders.
             string parsedExtension = ImageHelpers.Instance.GetExtension(path, querystring);
-            string encryptedName = (path).ToSHA1Fingerprint();
+            string hashedName = (path).ToSHA1Fingerprint();
 
-            return $"{encryptedName}.{(!string.IsNullOrWhiteSpace(parsedExtension) ? parsedExtension.Replace(".", string.Empty) : "jpg")}";
+            return $"{hashedName}.{parsedExtension.Replace(".", string.Empty)}";
         }
 
         /// <summary>
--- a/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
+++ b/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
@@ -29,6 +29,11 @@ namespace ImageProcessor.Web.Caching
         /// <returns>The <see cref="string"/></returns>
         public static string GetCachedImageFileName(string path)
         {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+
             int index = path.LastIndexOf("?", StringComparison.Ordinal);
 
             // Prevent overflow if path ends with a "?"
--- a/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
+++ b/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
@@ -1,0 +1,82 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="CachedFileHelper.cs" company="James Jackson-South">
+//   Copyright (c) James Jackson-South.
+//   Licensed under the Apache License, Version 2.0.
+// </copyright>
+// <summary>
+//   Provides helper methods to generate cached file names and paths.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace ImageProcessor.Web.Caching
+{
+    using System;
+    using System.IO;
+    using System.Linq;
+
+    using ImageProcessor.Web.Extensions;
+    using ImageProcessor.Web.Helpers;
+
+    /// <summary>
+    /// Provides helper methods to generate cached file names and paths.
+    /// </summary>
+    public static class CachedImageHelper
+    {
+        /// <summary>
+        /// Gets the cached file name from the given path.
+        /// </summary>
+        /// <param name="path">The path to the image. This can be the full path plus querystring.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedImageFileName(string path)
+        {
+            int index = path.LastIndexOf("?", StringComparison.Ordinal);
+
+            // Prevent overflow if path ends with a "?"
+            if (index == path.Length - 1)
+            {
+                index = 0;
+            }
+
+            return GetCachedImageFileName(path, index > 0 ? path.Substring(index + 1) : string.Empty);
+        }
+
+        /// <summary>
+        /// Gets the cached file name from the given path.
+        /// </summary>
+        /// <param name="path">The request path to the image. This can be the full path plus querystring.</param>
+        /// <param name="querystring">The request querystring.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedImageFileName(string path, string querystring)
+        {
+            // Use an sha1 hash of the full path including the querystring to create the image name.
+            // That name can also be used as a key for the cached image and we should be able to use
+            // The characters of that hash as sub-folders.
+            string parsedExtension = ImageHelpers.Instance.GetExtension(path, querystring);
+            string encryptedName = (path).ToSHA1Fingerprint();
+
+            return $"{encryptedName}.{(!string.IsNullOrWhiteSpace(parsedExtension) ? parsedExtension.Replace(".", string.Empty) : "jpg")}";
+        }
+
+        /// <summary>
+        /// Gets the path to the cached file.
+        /// </summary>
+        /// <param name="cachedFolderPath">The path to the cached folder, relative or absolute.</param>
+        /// <param name="cachedFileName">The cached file name.</param>
+        /// <param name="makeVirtual">Whether to reverse the slashes in the path.</param>
+        /// <param name="depth">How deep to nest the image files in folders.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedPath(string cachedFolderPath, string cachedFileName, bool makeVirtual, int depth = 6)
+        {
+            if (depth < 0)
+            {
+                depth = 0;
+            }
+
+            string pathFromKey = string.Join("\\", cachedFileName.ToCharArray().Take(Math.Min(Path.GetFileNameWithoutExtension(cachedFileName).Length, depth)));
+
+            return makeVirtual
+                ? Path.Combine(cachedFolderPath, pathFromKey, cachedFileName).Replace(@"\", "/")
+                : Path.Combine(cachedFolderPath, pathFromKey, cachedFileName);
+        }
+    }
+}
--- a/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
+++ b/src/ImageProcessor.Web/Caching/CachedImageHelper.cs
@@ -48,6 +48,16 @@ namespace ImageProcessor.Web.Caching
         /// <returns>The <see cref="string"/></returns>
         public static string GetCachedImageFileName(string path, string querystring)
         {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+
+            if (string.IsNullOrWhiteSpace(querystring))
+            {
+                throw new ArgumentNullException(nameof(querystring));
+            }
+
             // Use an sha1 hash of the full path including the querystring to create the image name.
             // That name can also be used as a key for the cached image and we should be able to use
             // The characters of that hash as sub-folders.
@@ -67,6 +77,16 @@ namespace ImageProcessor.Web.Caching
         /// <returns>The <see cref="string"/></returns>
         public static string GetCachedPath(string cachedFolderPath, string cachedFileName, bool makeVirtual, int depth = 6)
         {
+            if (string.IsNullOrWhiteSpace(cachedFolderPath))
+            {
+                throw new ArgumentNullException(nameof(cachedFolderPath));
+            }
+
+            if (string.IsNullOrWhiteSpace(cachedFileName))
+            {
+                throw new ArgumentNullException(nameof(cachedFileName));
+            }
+
             if (depth < 0)
             {
                 depth = 0;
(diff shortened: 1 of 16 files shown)


nul800sebastiaan:
I know we have plenty of experience with hashing now in Courier, so I'd love for @Shazwazza to have a look at this too.

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// <param name="path">The request path to the image. This can be the full path plus querystring.</param>
+        /// <param name="querystring">The request querystring.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedImageFileName(string path, string querystring)
+        {
+            // Use an sha1 hash of the full path including the querystring to create the image name.
+            // That name can also be used as a key for the cached image and we should be able to use
+            // The characters of that hash as sub-folders.
+            string parsedExtension = ImageHelpers.Instance.GetExtension(path, querystring);
+            string encryptedName = (path).ToSHA1Fingerprint();
```
Shazwazza: nit picky but this is hashed not encrypted :)

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// Provides helper methods to generate cached file names and paths.
+    /// </summary>
+    public static class CachedImageHelper
+    {
+        /// <summary>
+        /// Gets the cached file name from the given path.
+        /// </summary>
+        /// <param name="path">The path to the image. This can be the full path plus querystring.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedImageFileName(string path)
```
Shazwazza: Should always check params for null (or whitespace), r# does this for you

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+            return GetCachedImageFileName(path, index > 0 ? path.Substring(index + 1) : string.Empty);
+        }
+
+        /// <summary>
+        /// Gets the cached file name from the given path.
+        /// </summary>
+        /// <param name="path">The request path to the image. This can be the full path plus querystring.</param>
+        /// <param name="querystring">The request querystring.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedImageFileName(string path, string querystring)
```
Shazwazza: Should always check params for null (or whitespace), r# does this for you

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            // Use an sha1 hash of the full path including the querystring to create the image name.
+            // That name can also be used as a key for the cached image and we should be able to use
+            // The characters of that hash as sub-folders.
+            string parsedExtension = ImageHelpers.Instance.GetExtension(path, querystring);
```
Shazwazza: could the result of this be null?

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Gets the path to the cached file.
+        /// </summary>
+        /// <param name="cachedFolderPath">The path to the cached folder, relative or absolute.</param>
+        /// <param name="cachedFileName">The cached file name.</param>
+        /// <param name="makeVirtual">Whether to reverse the slashes in the path.</param>
+        /// <param name="depth">How deep to nest the image files in folders.</param>
+        /// <returns>The <see cref="string"/></returns>
+        public static string GetCachedPath(string cachedFolderPath, string cachedFileName, bool makeVirtual, int depth = 6)
```
Shazwazza: Should always check params for null (or whitespace), r# does this for you

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -207,46 +200,56 @@ public DiskCache(string requestPath, string fullPath, string querystring)
...
+                                 // have reduced the number of items below the maximum allowed.
+                                 // We'll cleanup any orphaned expired files though.
+                                 if (!this.IsExpired(fileInfo.CreationTimeUtc) && count <= MaxFilesCount - 1)
+                                 {
+                                     break;
+                                 }
+
+                                 // Remove from the cache and delete each CachedImage.
+                                 CacheIndexer.Remove(fileInfo.Name);
+                                 fileInfo.Delete();
```
Shazwazza: In this logic are you ever checking if the folder is empty and therefore deleting the folder too? This would be especially useful for FCN since ASP.NET's monitoring is per folder.

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -457,6 +460,49 @@ private static string GetValidatedAbsolutePath(string originalPath, out string v
...
+        /// <returns>The <see cref="bool"/></returns>
+        private async Task<bool> IsUpdatedAsync(DateTime creationDate)
+        {
+            bool isUpdated = false;
+
+            try
+            {
+                if (new Uri(this.RequestPath).IsFile)
+                {
+                    FileInfo imageFileInfo = new FileInfo(this.RequestPath);
```
Shazwazza: Instead of allocating a FileInfo instance which it seems you don't need you can achieve the same thing using the static System.IO.File methods to get the Exists and Utc date info from the file

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -457,6 +460,49 @@ private static string GetValidatedAbsolutePath(string originalPath, out string v
...
+                    request.Method = "HEAD";
+
+                    using (HttpWebResponse response = (HttpWebResponse)await request.GetResponseAsync())
+                    {
+                        isUpdated = response.LastModified.ToUniversalTime() > creationDate;
+                    }
+                }
+            }
+            catch
+            {
```
Shazwazza: That's a pretty big try/catch, where are you expecting this to go wrong? Maybe reduce the try catch or be more specific, perhaps log the exception too

src/ImageProcessor.Web/Caching/IImageCache.cs
```diff
@@ -41,6 +41,16 @@ public interface IImageCache
         int BrowserMaxDays { get; set; }
 
         /// <summary>
+        /// Gets or sets the maximum number folder levels to nest the cached images.
+        /// </summary>
+        int FolderDepth { get; set; }
```
Shazwazza: Adding to an interface will be a 'breaking' change, would either need to be documented since you aren't releasing a major version, or you could create an IImageCache2 interface to avoid the break

src/ImageProcessor.Web/Caching/DiskCache.cs
JimBobSquarePants: That's a good idea. Will have a look at implementing it.

src/ImageProcessor.Web/Caching/DiskCache.cs
JimBobSquarePants: Can you call refresh on the File? I recall using FileInfo so I could ensure the info wasn't cached by the OS. Guess I don't really need to do that.

src/ImageProcessor.Web/Caching/DiskCache.cs
JimBobSquarePants: There's no way for me to determine whether the request is from a remote url or a db or whatever IImageService implementation someone has created that isn't local. This was the cheapest way I could think of. Logging would probably just bloat the log when a custom service is used.

src/ImageProcessor.Web/Caching/IImageCache.cs
JimBobSquarePants: Yeah... I really didn't like doing this. Will have another stab at it.

nul800sebastiaan: Good point, people might be implementing this for blob storage purposes. Alternatively you could make this an extension method for now so it's not a breaking change (but then you also can't enforce implementing it). 
However, I can't find any examples of anybody actually implementing this interface, though it's possible it's in use privately.

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
JimBobSquarePants: :P True!

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
JimBobSquarePants: No it falls back to jpeg if it can't find anything.

nul800sebastiaan:
I would love to test these changes, how can I best test this? Set the cache time to a very low value? Anything else I should pay specific attention to?

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
JimBobSquarePants: Done

src/ImageProcessor.Web/Caching/DiskCache.cs
JimBobSquarePants: Done, An attempt will be made to delete the directory and it's parents up to the cache root but failure is handled and logged.

JimBobSquarePants:
@nul800sebastiaan Yeah, set the max Days to zero and you'll see it kick in. Sometimes the cache will fail to delete a directory as it's possible for a file to be added during enumeration. That's handled though. 

@Shazwazza I've pushed a bunch of changes and fixed the interface doodah also.

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
JimBobSquarePants: Done

src/ImageProcessor.Web/Caching/CachedImageHelper.cs
JimBobSquarePants: Done

nul800sebastiaan:
All good here, upgraded two live sites suffering from high CPU problems today, tested with setting cache `mayDays` to `0` helped clean up files/folders. If @Shazwazza is happy with the rest of the changes then we'd love a new build on NuGet soon! :-)
FYI: we're building 7.5.8 with this and some other changes and would like to do internal testing as soon as possible.

ZNS:
Haven't had the time to check the code, but it sounds like these changes really should help. I also like the new settings for depth and the possibility to turn trimming off. I'll probably turn it off for my sites and use a scheduled task to clean the cache. Not that I don't trust the solution, I just think I'm better off without the overhead of trimming the cache altogether.

Also if the Umbraco guys are happy, I'm happy. They know their stuff :)

Shazwazza:
My last piece of advice would be to run the trimming on a background thread since it's still running this on a request thread - although it is debounced this still means that the request won't actually end until the trimming is done. There's good and bad ways to do this... The super easy way is to use:

`ThreadPool.QueueUserWorkItem` which is a set and forget operation, however this is not web aware and you'd want to terminate the cleanup process when the app domain is shutting down otherwise it may not terminate quickly if this process is still running in the background. You could make this web aware if you implement IRegisteredObject in the class that is performing the cleanup.

Otherwise in .Net 4.5.2 you can use `HostingEnvironment.QueueBackgroundWorkItem` which is web aware but this is web aware in the fact that it will prevent the app domain from shutting down until the background job is done (or 30 seconds pass) which is sort of the opposite of what you want, but the good news is that this passes in a CancellationToken so you can just cancel the cleanup operation if it's been canceled and the app domain will stop waiting.

JimBobSquarePants:
I think I'll have to implement `IRegisteredObject ` to do this. I can't bump the minimum supported platform number from NET4.5. I'll have a look at it tonight.

Shazwazza:
Cool, you can also use Task.Run which essentially makes a process run on a threadpool thread too. If you do that you could instead make the class that calls Task.Run IRegisteredObject and then pass in a CancelationToken to Task.Run and when the IRegisteredObject methods fire you can cancel the CancelationTokenSource and then in the code that is running in Task.Run check against if the token is cancelled. 

Lots of ways to skin a cat :P

Shazwazza:
One other thing to know about running this stuff in background tasks is that any async/await is more or less irrelevant on a background thread - since it's already on a background thread. I might have time to help with this later tonight but let me know if you get to it before me

ZNS:
Might be able to shave some time off the process by optimizing file enumeration? 
http://stackoverflow.com/questions/17756042/improve-the-performance-for-enumerating-files-and-folders-using-net/17756314#17756314

nul800sebastiaan:
@JimBobSquarePants The suggestions here are great but we're getting more signals now from people who have high CPU problems, if you don't have time to work on these optimizations yet then it would be great to have the current state out as a new NuGet while these optimizations are being put in. People are currently quite stuck with this unless they downgrade to an older version. Any chance you can release the current state soon?

JimBobSquarePants:
I'll do a release tonight. The improvements here are vast already. 

Maybe I can slot the background thread stuff into v4.8.1.

nul800sebastiaan:
Awesome, yeps, that would be perfect! Asked Shan and he agrees that the extra improvements are not critical. Thanks a bunch James!

nul800sebastiaan:
@JimBobSquarePants 🥇 👍 
Get Shan to buy you another beer mate!

PR #543, opened 2017-01-18
Extensionless caching

ZNS:
Having diskcache outside of webroot does not work with extensionless URLs. This should fix that.

### DIFF
--- a/src/ImageProcessor.Web/Caching/DiskCache.cs
+++ b/src/ImageProcessor.Web/Caching/DiskCache.cs
@@ -296,7 +296,7 @@ namespace ImageProcessor.Web.Caching
                     // We need to manually write out the Content Type header here, the TransmitFile does not do this for you
                     // whereas the RewritePath actually does.
                     // https://github.com/JimBobSquarePants/ImageProcessor/issues/529
-                    string extension = Helpers.ImageHelpers.Instance.GetExtension(this.FullPath, this.Querystring);
+                    string extension = Path.GetExtension(this.CachedPath);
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Are you sure? I fall back to jpeg here.

https://github.com/JimBobSquarePants/ImageProcessor/blob/46d0a5241da1e3c55534dbd9da67ba4fd244daf3/src/ImageProcessor.Web/Helpers/ImageHelpers.cs#L139

ZNS:
I have a special case where I need to handle an url like "/{imageid}/{width}/{height}/" I've implemented this using the ValidatingRequest event which works. And yes, the fallback works but if the image is not a jpg it will obviously fail. Also I think it makes more sense getting the extension from the actual file being transfered than from the URL?

ZNS:
Come to think of it I haven't tested this with anything but jpg. Perhaps there are other issues when using an extensionless url. I'll test that.

JimBobSquarePants:
The cached file could have the wrong extension because the fallback has always been jpg. Yeah please do, we should really be reading the stream to return the correct content type to the browser which will essentially ignore the extension.

ZNS:
You're right. The cached image always gets the ".jpg" extension, even if it's a gif or png. It seems a workaround for this is to pass "format=xxxxx" with the querystring.

JimBobSquarePants:
@ZNS What exactly happens when the server tries to transmit the file? 

I was doing some testing and the StaticFileHandler for internally cached images will also return the wrong mime type (image/jpeg) if it falls back to jpeg and the extension is wrong. All the browsers I've tested against doint seem to care though.

Issue #544, opened 2017-01-18
content-length not sent when storing images outside of web root


alindgren:
Need to set the content-length when using `TransmitFile()` method.

alindgren:
I'm testing a fix now - https://github.com/alindgren/ImageProcessor/commit/a11b32a2d13550793f5a873f4b8198f88d375537

JimBobSquarePants:
Thanks... Disappointed I missed this.

alindgren:
@JimBobSquarePants I think several people were looking at this code and missed it. But this is OSS in action.  PR just submitted :)

JimBobSquarePants:
Fixed in ImageProcessor.Web v4.8.1

PR #545, opened 2017-01-18
add content-length header when using TransmitFile method

alindgren:
fixes issue https://github.com/JimBobSquarePants/ImageProcessor/issues/544


JimBobSquarePants:
Aye that should do it. Thanks!

Issue #537, opened 2017-01-12
Check for supportedImageFormat should use more than 4 bytes


dampee:
> the PDF spec says the %PDF-1.x only needs to be in the first 1024 bytes and not the first 4...   ([from stackoverflow](http://stackoverflow.com/a/35929093/97615))

This is an unsupported scenario for ImageProcessor because we check the "Supported File Format" by [taking the first 4 bytes of a file](https://github.com/JimBobSquarePants/ImageProcessor/blob/develop/src/ImageProcessor/Imaging/Formats/FormatUtilities.cs#L49). 
Secondly, if a header of 5 or more bytes is provided by the new ImageFormat, the `GetFormat` method in the `FormatUtilities` class will not complain, but just return null

I see a few solutions:

1. If the supplied header is longer than four bytes, just throw an exception telling this is not supported.  I guess this is not the best solution because you would never be able to correctly implement a "PDF format". And possibly other future formats...
2. Read the maximum number of bytes possibly needed in Imaging.Formats.FormatUtilities ([line 41](https://github.com/JimBobSquarePants/ImageProcessor/blob/develop/src/ImageProcessor/Imaging/Formats/FormatUtilities.cs#L49)): 

    var numberOfBytesToRead = supportedImageFormats.Max(f => f.FileHeaders.Max(h=>h.Length));
    byte[] buffer = new byte[numberOfBytesToRead];

3. or only use the four bytes supplied in the header ([line 59](https://github.com/JimBobSquarePants/ImageProcessor/blob/develop/src/ImageProcessor/Imaging/Formats/FormatUtilities.cs#L59)) of the new fileformat

    if (header.**Take(4)**.SequenceEqual(buffer.Take(header.Length)))

JimBobSquarePants:
Thanks for this.

Ideally we should have variable length support using the maximum length (we do that in ImageSharp). Would you have the chance to do a PR?

Cheers

James

dampee:
Sure.

dampee:
Untested, no pull request yet.

JimBobSquarePants:
Fixed in ImageProcessor v2.5.2

PR #546, opened 2017-01-18
support file header with variable length (#537)

dampee:
This will also allow to have [jpegs with longer fileheaders](https://github.com/JimBobSquarePants/ImageProcessor/blob/develop/src/ImageProcessor/Imaging/Formats/JpegFormat.cs#L31)


JimBobSquarePants:
Ah great stuff. Looks good thanks!

PR #553, opened 2017-01-30
Trim cache runs on a background thread with a sliding expiration

Shazwazza:
### Prerequisites

### Description

This moves all cache trimming to execute on background threads. Currently when cache trimming occurs it runs on the request thread and will block the request thread (even though the image response has been sent) until the cache trimming is done. In some cases this can be a lengthy operation and also causes high CPU spikes when there are a lot of folders. 

To fix this:

* Cache trimming will run on a background thread
* The background thread will be a timer thread
* Trimming will only execute when it's 'quiet', which means that there are no currently executing IP requests to generate new images. Each time `ScheduleTrimCache` is called it will postpone the trimming by 2 minutes. There is a maximum timeout however so after it's been waiting for 30 minutes even if there are current IP requests coming in for new images the cache trimming will still execute
* The cache trimming is web aware with IRegisteredObject which will cancel the cancellation token which is passed in to all logic that performs the trimming - this same token also gets passed to the Azure Blob storage APIs so that if the app is shutting down it does not hold up the app shutdown
* Cancellation tokens are used in all logic of trimming and are checked in each loop so that if the app is shutting down the trimming will stop
* The DiskCache trimming has a perf fix before if there are thousands of folders, when it was checking for empty folders, it would re-enumerate the entire hierarchy for each folder/files


JimBobSquarePants:
@Shazwazza Well I'm glad you knew how to do this! 😄 I'll download and do some testing asap.

JimBobSquarePants:
I've merged the changes into dev. Tweaked the cleanup code in DiskCache slightly also reversing the DirectoryInfo order to search bottom up. This results in a more effective, less intensive cleanup. 

Really happy with the cleanup strategy now. Thanks for your help!

PR #570, opened 2017-03-30
Fix services cannot have same type but different settings

lars-erik:
Given the following security.config:

    <?xml version="1.0" encoding="utf-8"?>
    <security>
      <services>

        <service prefix="media/" name="CloudImageService" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
          <settings>
            <setting key="Container" value="media" />
            <setting key="MaxBytes" value="8194304" />
            <setting key="Timeout" value="30000" />
            <setting key="Host" value="http://[customer].blob.core.windows.net/" />
          </settings>
        </service>

        <service prefix="360-views/" name="360ViewsService" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
          <settings>
            <setting key="Container" value="360-views" />
            <setting key="MaxBytes" value="8194304" />
            <setting key="Timeout" value="30000" />
            <setting key="Host" value="https://[customer].blob.core.windows.net/" />
          </settings>
        </service>
    
      </services>
    </security>

When generating URLs, the 360ViewsService will use the Container from the CloudImageService settings. This is because settings is fetched based on Type name instead of service configuration name. Of course resulting in a 404.

This fix assigns the name in the config to a new name property on services, enabling the code to fetch their respective settings. If name isn't given, the type name is used.

I had to fix this for a project I'm working on and I'm sorry I don't have the time to add a test for it. I imagine the test in question must be missing, though. Soz. Shout out if it's a deal-breaker. ;)

### DIFF
--- a/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
+++ b/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
@@ -304,43 +304,16 @@ namespace ImageProcessor.Web.Configuration
                 IImageService imageService = Activator.CreateInstance(type) as IImageService;
                 if (imageService != null)
                 {
-                    if (!string.IsNullOrWhiteSpace(config.Prefix))
-                    {
-                        imageService.Prefix = config.Prefix;
-                    }
-                    if (!string.IsNullOrWhiteSpace(config.Name))
-                        imageService.Name = config.Name;
-                    else
-                        imageService.Name = type.Name;
+                    var name = config.Name ?? imageService.GetType().Name;
+                    imageService.Prefix = config.Prefix;
+                    imageService.Settings = GetServiceSettings(name);
+                    imageService.WhiteList = GetServiceWhitelist(name);
                 }
 
 
                 this.ImageServices.Add(imageService);
             }
 
-            // Add the available settings.
-            foreach (IImageService service in this.ImageServices)
-            {
-                string name = service.Name;
-                Dictionary<string, string> settings = this.GetServiceSettings(name);
-                if (settings.Any())
-                {
-                    service.Settings = settings;
-                }
-                else if (service.Settings == null)
-                {
-                    // I've noticed some developers are not initializing 
-                    // the settings in their implentations.
-                    service.Settings = new Dictionary<string, string>();
-                }
-
-                Uri[] whitelist = this.GetServiceWhitelist(name);
-
-                if (whitelist.Any())
-                {
-                    service.WhiteList = this.GetServiceWhitelist(name);
-                }
-            }
         }
 
         /// <summary>
--- a/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
+++ b/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
@@ -302,13 +302,18 @@ namespace ImageProcessor.Web.Configuration
                 }
 
                 IImageService imageService = Activator.CreateInstance(type) as IImageService;
-                if (!string.IsNullOrWhiteSpace(config.Prefix))
+                if (imageService != null)
                 {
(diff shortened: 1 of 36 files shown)


lars-erik:
OK, I refactored the config instead of breaking the interface. Now given the following config:

    <?xml version="1.0" encoding="utf-8"?>
    <security>
      <services>

        <service prefix="media/" name="CloudImageService" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
          <settings>
            <setting key="Container" value="media" />
            <setting key="MaxBytes" value="8194304" />
            <setting key="Timeout" value="30000" />
            <setting key="Host" value="http://[customer].blob.core.windows.net/" />
          </settings>
        </service>

        <service prefix="360-views/" name="360ViewsService" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
          <settings>
            <setting key="Container" value="360-views" />
            <setting key="MaxBytes" value="8194304" />
            <setting key="Timeout" value="60000" />
            <setting key="Host" value="https://[customer].blob.core.windows.net/" />
          </settings>
        </service>

        <service prefix="other/" name="other" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
          <whitelist>
            <add url="asdf"/>
            <add url="aerwt"/>
          </whitelist>
        </service>

        <service prefix="other2/" name="other2" type="ImageProcessor.Web.Services.CloudImageService, ImageProcessor.Web">
        </service>
    
      </services>
    </security>

The services will be set up as follows:

    [
        {
            "Prefix": "media/",
            "IsFileLocalService": false,
            "Settings": {
                "Container": "media",
                "MaxBytes": "8194304",
                "Timeout": "30000",
                "Host": "http://[customer].blob.core.windows.net/"
            },
            "WhiteList": []
        },
        {
            "Prefix": "360-views/",
            "IsFileLocalService": false,
            "Settings": {
                "Container": "360-views",
                "MaxBytes": "8194304",
                "Timeout": "60000",
                "Host": "https://[customer].blob.core.windows.net/"
            },
            "WhiteList": []
        },
        {
            "Prefix": "other/",
            "IsFileLocalService": false,
            "Settings": {},
            "WhiteList": [
                "asdf",
                "aerwt"
            ]
        },
        {
            "Prefix": "other2/",
            "IsFileLocalService": false,
            "Settings": {},
            "WhiteList": []
        }
    ]

(serialized from memory while debugging, so there's your test. :) )

Quite a bit of the logic now gone was reduntant since the config factory methods already return empty collections. Might as well just set them instead of creating new empty collections.

src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
```diff
@@ -302,40 +302,18 @@ private void LoadImageServices()
...
 
                 IImageService imageService = Activator.CreateInstance(type) as IImageService;
-                if (!string.IsNullOrWhiteSpace(config.Prefix))
+                if (imageService != null)
                 {
-                    if (imageService != null)
-                    {
-                        imageService.Prefix = config.Prefix;
-                    }
+                    var name = config.Name ?? imageService.GetType().Name;
```
lars-erik: Might not need the GetType().Name thing. Name is required on the config element, so we'll always have one.

JimBobSquarePants: Haha! If you've solved this in such a simple manner I think I may owe you a beer. I'll dl and have a good look asap. 👍

lars-erik: CG18 :)

JimBobSquarePants: It's a date 😄

PR #580, opened 2017-04-10
fix gif add watermark font bug

g7rhythm:
if gif image set watermark font will fail .
because FontFamily object is Dispose .

Thanks for contributing to ImageProcessor!

### DIFF
--- a/src/ImageProcessor/Processors/Watermark.cs
+++ b/src/ImageProcessor/Processors/Watermark.cs
@@ -206,9 +206,10 @@ namespace ImageProcessor.Processors
         {
             try
             {
-                using (fontFamily)
+				var f = new FontFamily(fontFamily.Name);
+                using (f)
                 {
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Hi @g7rhythm Can you expand upon this please. 

This has nothing to do with gif files rather trying to use a fontfamily in a textlayer more than once. 

Can you please expand with a complete code sample for me to replicate with?

Cheers

James

JimBobSquarePants:
I think I should make TextLayer implement IDisposable and include your PR. Thanks for spotting this and contributing! 👍

g7rhythm:
if gif file have many frame and use fontfamily add a watermark.
just first frame can show the fontStyle. other frame will use default font to show the text,  because FontFamily object is Dispose .

JimBobSquarePants:
Thanks. I've already merged your PR and will release a new version soon.

Issue #255, opened 2015-11-26
Trojan Vundo.IB detected


perosb:
FYI, probably a false positive, Windows Defender detect a Trojan:Win32 Vundo.IB inte ImageProcessor.Web.dll

```
The following error occurred: Error code 0x80508023. The program could not find the malware and other potentially unwanted software on this computer. 

Category: Trojan

Description: This program displays advertisements and may be difficult to remove.

Recommended action: Remove this software immediately.

Items: 
file:C:\_Projects\web\p\App_Data\NuGetBackup\20151123-205238\bin\ImageProcessor.Web.dll
```

JimBobSquarePants:
Yeah.. See #162 Defender can't recognise a png optimizer when it sees one. False positives in two libraries now.

Issue #162, opened 2015-05-04
Getting a malware trojan warning from our local machines for ImageProcessor.dll


kgiszewski:
Should we be concerned?

![screen shot 2015-05-04 at 9 51 31 am](https://cloud.githubusercontent.com/assets/4653907/7455041/f9c5170a-f246-11e4-9257-6c30464591cf.png)

cosmo0:
Well, it can come from pretty much anywhere.

ImageProcessor itself does not contain malicious code (that's a benefit of open source, you can check it out), but we cannot guarantee that the place that you got your DLL from haven't added code on top of it.

JimBobSquarePants:
It certainly shouldn't, however I want to double check that nothing has somehow got piggy backed. What antivirus is that? Could you run the file though https://www.virustotal.com/

JimBobSquarePants:
I think your server is compromised. Here's a scan of the binary from the latest package. https://www.virustotal.com/en/file/96ac35fd1eba8c0414ed8b0a051d679c52d8ea5310e0377a62fe303228016eda/analysis/1430752679/

kgiszewski:
Right so, as an update...

Our campus-wide malware definitions were updates on the 1st and that's when we started getting the notification (we use MalwareBytes).  We get it on any machine using NuGet as our source.

I'm not at all concerned with the source code and since our checksum matches, I'll have to just ignore it for now.

Closing, but thought I'd ask :)

JimBobSquarePants:
Raising it was definitely the right thing to do. Thanks.

JimBobSquarePants:
Just to note. I will no longer be using png.out in future versions to post process images. That is the file which has been triggering the antivirus.

nul800sebastiaan:
After receiving 3 questions about this in a week, it seems like windows defender is now detecting this as a threat more aggressively than before. This is obviously not a good experience when you try out Umbraco for the first time, and I can't imagine it's a good experience for people checking out ImageProcessor for the first time. 
On top of that, there's now been a few reports of the post processing tools not working on some hosts (#268 and #279). And it might be causing performance problems for some people with large amounts of images, which are unconfirmed, and I've not looked into this at all so it might not be related at all: https://our.umbraco.org/forum/umbraco-7/using-umbraco-7/73814-umbraco-728-upgrade-to-733-caused-memory-issues-tracked-down-to-upgrade-of-imageprocessor

I think we have a few options:
[...]

PR #581, opened 2017-04-10
Added optipng to the PNG post processor.

gerbenvd:
Aded optipng to the PNG post processor. This seems to have better results when optimizing with Google PageSpeed Insights

### Prerequisites

### Description
I've added optiPNG to the list of post processors that is used when compressing PNG files.


JimBobSquarePants:
Hi @gerbenvd 

Thanks for taking the time to contribute to the project with your PR. 

Unfortunately there is a long standing issue with optipng that causes Windows Defender to return a false positive.
 
https://github.com/madskristensen/WebEssentials2013/issues/541

I was actually using the program before I received the following issues.

https://github.com/JimBobSquarePants/ImageProcessor/issues/162

https://github.com/JimBobSquarePants/ImageProcessor/issues/255

If we can establish that this is no longer the case then I could accept the PR. Until then I sadly cannot.

Cheers

James

gerbenvd:
Hi Jim!

I understand your considerations. In hindsight, I could have looked better on the open issues, my bad!

Did you consider adding another PNG optimizer? Below you find some references that do a comparision of the various progams out there:

http://www.olegkikin.com/png_optimizers/
https://css-ig.net/png-test-corpus

Would it be an option to make the compression pipe-line configurable, so that you can set which compression programs will be used by means of a configuration file?

JimBobSquarePants:
@gerbenvd I certainly would consider it yes. 

We should probably copy the approach used by [ImageOptimizer](https://github.com/madskristensen/ImageOptimizer/tree/master/src) and improve our compression throughout, except don't use Google guetzli anywhere since the memory requirements are too high for a webserver.

Is that something you could dedicate time to do?

Cheers

James

JimBobSquarePants:
Found the time to do it myself so I've actually just replaced pngout with truepng. This makes any PageSpeed issues go away. 

https://github.com/JimBobSquarePants/ImageProcessor/commit/871fa4a06b8784e85ca4f75cac8566387dd41dfb

gerbenvd:
You fixed it before I got time to respond :) Thanks!!

JimBobSquarePants:
No worries 😄

PR #582, opened 2017-04-23
use ManualResetEventSlim instead of using Thread.Sleep in a loop

tmaczynski:
This is a minor change which eliminates anti-pattern of using Thread.Sleep() in a loop. Using Thread.Sleep() in a loop is inefficient because of unnecessary context switches and unnecessary delays that it causes.

I've found this repo by using [Google BigQuery](https://cloud.google.com/bigquery/public-data/github) when I searched if [my code analyzer](https://github.com/DotNetAnalyzers/AsyncUsageAnalyzers/pull/50) is useful.

PS. I recommend viewing this PR is a in a tool that handles changes of indentation well (e.g. git plugin for VS2017).

Prerequisites:

- [x] I have verified that there are no overlapping pull-requests open
- [x] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules 👮.

### DIFF
--- a/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
+++ b/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
@@ -100,50 +100,42 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
                 return null;
             }
 
-            int elapsedTime = 0;
-            bool eventHandled = false;
-            Process process = null;
-            try
-            {
-                process = new Process
+            using (var processingFinished = new ManualResetEventSlim(false))
+            { 
+                Process process = null;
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Thanks for this appreciate it! 👍 

BTW that analyzer is very clever. Hopefully you get the appropriate feedback soon!

tmaczynski:
Thanks:)

To be clear: My analyzer provides code fixes only if `Thread.Sleep` is used in `async` code, this change was found by my analyzer and I changed the code myself.

JimBobSquarePants:
Ah cool, Still clever stuff. Something I'd like to look into one day.

PR #584, opened 2017-05-02
Add support for regular expression as the ImageService prefix.

CameronWills:
### Prerequisites

- [?] I have provided test coverage for my change (where applicable)

### Description
In some cases its useful to have the same ImageService instance handle multiple prefixes, or better still, a pattern for the prefix. This is particularly relevant for some .net CMSs that use different paths for different media buckets. This is becoming a pain in my current project - having five `<service ... >` elements in the security.config mapped to the same IImageService type.

Unfortunately, this change is breaking - as it adds another property to IImageService
`public bool IsRegex { get; set; }`

Thoughts?

### DIFF
--- a/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
+++ b/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
@@ -826,7 +826,10 @@ namespace ImageProcessor.Web.HttpModules
             foreach (IImageService service in services)
             {
                 string key = service.Prefix;
-                if (!string.IsNullOrWhiteSpace(key) && path.StartsWith(key, StringComparison.InvariantCultureIgnoreCase))
+                if (!string.IsNullOrWhiteSpace(key) 
+                    && (service.IsRegex 
+                        ? Regex.IsMatch(path, key, RegexOptions.IgnoreCase) 
+                        : path.StartsWith(key, StringComparison.InvariantCultureIgnoreCase)))
                 {
                     return service;
                 }
--- a/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
+++ b/src/ImageProcessor.Web/Configuration/ImageProcessorConfiguration.cs
@@ -307,6 +307,7 @@ namespace ImageProcessor.Web.Configuration
                 {
                     string name = config.Name ?? imageService.GetType().Name;
                     imageService.Prefix = config.Prefix;
+                    imageService.IsRegex = config.IsRegex;
                     imageService.Settings = this.GetServiceSettings(name);
                     imageService.WhiteList = this.GetServiceWhitelist(name);
                 }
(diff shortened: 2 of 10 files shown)


src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
```diff
@@ -826,7 +826,10 @@ private IImageService GetImageServiceForRequest(string url, string applicationPa
             foreach (IImageService service in services)
             {
                 string key = service.Prefix;
-                if (!string.IsNullOrWhiteSpace(key) && path.StartsWith(key, StringComparison.InvariantCultureIgnoreCase))
+                if (!string.IsNullOrWhiteSpace(key) 
+                    && (service.IsRegex
```
JimBobSquarePants: How about a logical or here instead? That way you could prevent altering the interface.

```
&& (Regex.IsMatch(path, key, RegexOptions.IgnoreCase) 
        || path.StartsWith(key, StringComparison.InvariantCultureIgnoreCase)
```

CameronWills: Yep happy to make this change.. as long as you don't see any problems with existing prefix values containing any regex special characters ? Regex special characters are `[\^$.|?*+()`
I would think the dot (.) character is the most risky of these?

E.g. requests like `http://www.domain.com/remote-axd?...` would be matched with by the RemoteImageService: `<service prefix="remote.axd" name="RemoteImageService" ...>`

JimBobSquarePants: `Regex.IsMatch(path, Regex.Escape(key), RegexOptions.IgnoreCase)`

JimBobSquarePants: ⬆️ That should do it.

CameronWills: But wouldn't escaping the prefix mean that you cannot actually use a regular expression as the prefix? Effectively converts special characters in the regex to a literal characters?
For example, an ImageService with regular expression for the prefix `prefix="firstprefix|secondprefix"` to match these urls:
`http://www.domain.com/firstprefix/image.jpg`
`http://www.domain.com/secondprefix/image.jpg`

Regex escaping this prefix will give `firstprefix\|secondprefix`

Both sides of this OR expression would be false?
```(Regex.IsMatch(path, Regex.Escape(key), RegexOptions.IgnoreCase) 
        || path.StartsWith(key, StringComparison.InvariantCultureIgnoreCase)```

JimBobSquarePants: Bugger, you're correct! It' looks like we'll have to delay this then as I cannot add the property until I release V5. Shame!

CameronWills: Yeah fair enough. Do you have a roughly planned ETA on ImageProcessor.Web V5?

JimBobSquarePants: After I get [ImageSharp](https://github.com/JimBobSquarePants/ImageSharp) V1 out of the way most likely.  Could be as late as the end of the year.

PR #613, opened 2017-07-03
Adding in support for private blob storage in Azure

mtcomputergirl:
### Prerequisites

- [X] I have provided test coverage for my change (where applicable)

### Description

Added in a config value for an Azure connection string, added in the Azure Storage libraries. Added code to detect if a connection string is available in the config, and if so, use it to connect to the blob using the Azure Storage library. The code uses a config setup aligning with the Azure FSP.


JimBobSquarePants:
Thanks for this. I'll do a review as soon as possible! 👍

JimBobSquarePants:
@mtcomputergirl I meant to comment on this ages ago but forgot, please accept my apologies. 

Unfortunately the PR, as it stands can not be accepted. I can't have any references to Azure in the main ImageProcessor.Web repo.

I think perhaps this code would be better suited in a custom `IImageService` implementation

Thanks again for your interest.

James

Issue #623, opened 2017-08-29
Exception "Value of '-241' is not valid for 'alpha' when converting and saving image


bunnu:
### Description
I'm converting multiple files from multiple formats to indexed png. Some particular files can't be converted, because ImageProcessor throws following exception:
`Value of '-241' is not valid for 'alpha'. 'alpha' should be greater than or equal to 0 and less than or equal to 255.`

Stacktrace:
``` at ImageProcessor.Imaging.Quantizers.WuQuantizer.WuQuantizerBase.Quantize(Image source, Int32 alphaThreshold, Int32 alphaFader, Histogram histogram, Int32 maxColors)
at ImageProcessor.Imaging.Quantizers.WuQuantizer.WuQuantizerBase.Quantize(Image source)
at ImageProcessor.Imaging.Formats.PngFormat.Save(String path, Image image, Int64 bitDepth)
at ImageProcessor.ImageFactory.Save(String filePath)
```

### Steps to Reproduce
Sample code:
```c#
static void Main(string[] args)
{
    Thread.CurrentThread.CurrentUICulture = new CultureInfo("en-US");
    try
    {
        using (ImageFactory imageFactory = new ImageFactory())
        {
            var directory = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
            imageFactory.Load($"{directory}\\input.jpeg")
                        .Format(new PngFormat() { IsIndexed = true })
                        .Save($"{directory}\\output.png");
        }
        Console.WriteLine("Success.");
    }
    catch(Exception ex)
    {
        Console.WriteLine(ex.Message);
    };

    Console.ReadLine();
}
````

A sample input-file, which causes this exception is [input.jpeg](https://user-images.githubusercontent.com/5147219/29809169-6dc404a4-8c9b-11e7-8510-48a9f75cbe4d.jpeg).

The criticised value isn't always -241. If you open the image in paint.net, add a few strokes and save it as jpeg you most probably get an exception with value -242.

### System Configuration

- ImageProcessor version: 2.5.4
- Other ImageProcessor packages and versions: 2.5.0, 2.4.0, 2.3.0, 2.2.0
- Environment (Operating system, version and so on): Win 10 Pro 1703 x64
- .NET Framework version: 4.6.1

JimBobSquarePants:
Fixed in 2.5.5

PR #624, opened 2017-08-31
fix: value of '-xyz' is not valid for 'alpha'

bunnu:
### Description
The fix is taken from philjones' PR in https://github.com/drewnoakes/nquant/pull/3.
This fixes the issue #623


JimBobSquarePants:
Aw brilliant! Thanks! 👍

PR #628, opened 2017-09-19
Added background processor

ronaldbarendse:
The new Background processor is a combination of the existing BackgroundColor and Overlay processors.

Works perfectly when a background color is not sufficient (e.g. the background should have a gradient) and the original image has transparency (PNG): instead of adding the gradient background to the original PNG in the webpage itself, ImageProcessor can combine these and change the format to JPEG. This results in 1 less HTTP request (for the seperate background image) and smaller images (since JPEG is almost always smaller than PNG).


JimBobSquarePants:
Hi, @ronaldbarendse, thanks for this.

Could you perhaps post an example image of the result so I can understand more clearly?

Cheers

James

ronaldbarendse:
I'll do my best to get an example online soon, but the result is a background image behind the processed image (similar like the background color).

Explained otherwise, this works like the overlay processor, but swaps the image layers (it adds an 'underlay' image).

JimBobSquarePants:
Great. Looking forward to it 👍

ronaldbarendse:
I've added an example to the `ImageProcessor.TestWebsite` project, so you can see the result of combining a background JPEG with a PNG source/input image.

I also added the config transforms, so new projects will have a default config for the processor (although it's disabled by default, just like the overlay processor).

JimBobSquarePants:
@ronaldbarendse Finally getting around to merging this in. Thanks for your patience! 👍

PR #636, opened 2017-11-07
Changed Debug.WriteLine to Trace.WriteLine

ronaldbarendse:
Allow logging messages without DEBUG constant or resorting to reflection using the built-in Trace class.

### DIFF
--- a/src/ImageProcessor/Common/Exceptions/Logging/DefaultLogger.cs
+++ b/src/ImageProcessor/Common/Exceptions/Logging/DefaultLogger.cs
@@ -12,37 +12,15 @@
 namespace ImageProcessor.Common.Exceptions
 {
     using System;
-    using System.Linq.Expressions;
-    using System.Reflection;
+    using System.Diagnostics;
     using System.Runtime.CompilerServices;
 
     /// <summary>
-    /// The default logger which logs messages to the debugger.
+    /// The default logger which logs messages to the trace listeners.
     /// </summary>
+    /// <seealso cref="ImageProcessor.Common.Exceptions.ILogger" />
     public class DefaultLogger : ILogger
     {
-        /// <summary>
-        /// The writeline delegate.
-        /// </summary>
-        private readonly Action<string> debugWriteLine;
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="DefaultLogger"/> class.
-        /// </summary>
-        public DefaultLogger()
-        {
-            // Runtime compile the Debug.WriteLine(string text) method so we can call it outside 
-            // debug compilation.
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Hi @ronaldbarendse 

Thanks for your input here.

I originally wrote the default logger this way to allow calling the method in release mode.by bypassing the restriction dynamically invoking the method at runtime using expression trees. 

See https://stackoverflow.com/a/23669216/427899

I specifically didn't want to have to use trace since it required creating a `TraceListener`.

However, I could never figure out where it would actually log to since it won't log to the debug window.

How would you envision this working? Would you define the listener in the consuming app defining where would it write to?

If you want to write your own logger using Trace or other it might be simpler to implement the  interface and register your implementation via the [`SetLogger`](https://github.com/JimBobSquarePants/ImageProcessor/blob/2704e38f062fb3c9267193129eea5cd363887208/src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs#L86) method in `ImageProcessorBootstrapper`

Cheers

James

ronaldbarendse:
The `Debug` and `Trace` classes both share the same listeners (from https://docs.microsoft.com/nl-nl/dotnet/api/system.diagnostics.trace#Remarks):

_The `Listeners` collection is shared by both the `Debug` and the `Trace` classes; adding a trace listener to either class adds the listener to both. By default, trace output is emitted using the `DefaultTraceListener` class._

The [`DefaultTraceListener`](https://docs.microsoft.com/nl-nl/dotnet/api/system.diagnostics.defaulttracelistener#Remarks) writes it output to the debugger output window, so the app doesn't need special configuration to use this logger. Having code that is compiled without the DEBUG preprocessor directive still logging the the debug window also seems wrong to me, especially when combined with reflection to do so.

The `ILogger` interface doesn't support different event types (Verbose, Information, Warning, Error), otherwise a combination could be made using `Debug.WriteLine`, `Trace.TraceInformation`, `Trace.TraceWarning` and `Trace.TraceError`... Looking at the `ILogger.Log` calls, most are warning or error messages though ([Umbraco ](https://github.com/umbraco/Umbraco-CMS/blob/d04510f5f2882ec837fd7310bd977ac8abbe4b62/src/Umbraco.Core/Logging/ImageProcessorLogger.cs) also logs everything as an error), so maybe the implementation would be more correct if it used `Trace.TraceError` instead. Hmmm...

But just like you suggest, creating a new implementation is always possible and should probably be done when using the code in production anyways...

JimBobSquarePants:
@ronaldbarendse I'm gonna go with this. Re-reading the docs it make a much more sensible default. Thanks!

PR #640, opened 2017-11-22
Allow `CachedRewritePathExpiry` to be passed as a config variable

owenblacker:
### Prerequisites

- [ ] ~~I have provided test coverage for my change (where applicable)~~

### Description
I've added a new optional config property `ImageCacheSection.CachedRewritePathExpiry` and passed this in to `CacheIndexer.Add` with a default value matching the currently-hardcoded expiry time of 1 minute. This necessitates a new property `ImageCacheBase.CachedPathExpiry`.

### DIFF
--- a/src/ImageProcessor.Web/Caching/CacheIndexer.cs
+++ b/src/ImageProcessor.Web/Caching/CacheIndexer.cs
@@ -56,18 +56,21 @@ namespace ImageProcessor.Web.Caching
         }
 
         /// <summary>
-        /// Adds the specified key and value to the dictionary or returns the value if it exists.
+        /// Adds a <see cref="cachedImage"/> to the cache.
         /// </summary>
         /// <param name="cachedImage">
         /// The cached image to add.
         /// </param>
+        /// <param name="expiry">
+        /// The number of minutes to cache the image, defaults to 1.
+        /// </param>
         /// <returns>
         /// The value of the item to add or get.
         /// </returns>
-        public static CachedImage Add(CachedImage cachedImage)
+        public static CachedImage Add(CachedImage cachedImage, int expiry = 1)
         {
-            // Add the CachedImage with a sliding expiration of 1 minutes.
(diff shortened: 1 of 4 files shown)


JimBobSquarePants:
Hey @owenblacker  Thanks for this!

I never considered this something I would configure as I didn't want to use up a ton of memory, what was your use case? Issues with the disk cache?

Cheers

James

owenblacker:
Hey @JimBobSquarePants, no problem.

There's a dozen sites we maintain for a client that are built by us, but the hosting infrastructure is the domain of another supplier and they wanted a longer cache duration (of 10 minutes).

We've been running with this change for about 2 years now without a problem and it defaults to the value that's hard-coded in the `master` branch, so it should be pretty safe to integrate :)

Thanks!

Owen

JimBobSquarePants:
@owenblacker Awesome! Well let's get this merged in then. Thanks! 👍

dlemstra:
@JimBobSquarePants Wouldn't it be better to use a `TimeSpan` here instead of an `int`? This will make it more readable.

owenblacker:
@dlemstra Good catch; thank you!

I used an `int` because I was transposing a value from our Web.config, so it couldn't be strongly typed and, tbh, it didn't occur to me that I could use a `struct` like `TimeSpan` as an [optional argument](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments) and that [`TimeSpanValidatorAttribute`](https://msdn.microsoft.com/en-us/library/system.configuration.timespanvalidatorattribute(v=vs.110).aspx) makes them actually quite easy to specify like `0:10:0`. In retrospect, I probably should have done exactly  that, but that wouldn't have matched the private fork we'd been running on our client site for a couple of years so I just didn't think of it.

This feels like a perfect case for a method overload, though. The property in `ImageCacheSection` could just as easily use a `TimeSpanValidatorAttribute`, so I'll chat with the dev lead (who has taken over that project from me) about migrating to do that in our codebase.

owenblacker:
I've made this change and submitted it as a new pull request #644.

PR #641, opened 2017-11-28
Set 'prefix' config value for CloudImageService

SteveVaneeckhout:
### Prerequisites

### Description

This commit sets a value for the prefix attribute in the CloudImageService-service element security.config file. The attribute is necessary for ImageProcessor to work and because files in Umbraco are always saved in the same Media directory we can safely provide a default value for this setting.


JimBobSquarePants:
Hi Steve, 

Thanks for contributing but unfortunately I cannot accept this PR.

`Prefix` is an optional property used to filter requests, defaulting to `string.Empty` . It's perfectly acceptable for someone to use the service to capture requests without it. 

Additionally, `media/` is specific to Umbraco. I'm not sure where this confusion comes from but the library is not, and has never been an Umbraco dependent library (They merely consume it).

Cheers

James

PR #644, opened 2018-01-03
Amend `CachedRewritePathExpiry` to be passed as a `TimeSpan` (instead of `int`)

owenblacker:
### Prerequisites
- [x ] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [ ] ~~I have provided test coverage for my change (where applicable)~~

### Description
Amend `CachedRewritePathExpiry` to be passed as a `TimeSpan` (instead of `int`) as suggested by @dlemstra in pull request #640.

I've made the `TimeSpan` be the default, but added an overload to `CacheIndexer.Add` that takes an `int`, in case anyone has already started using the code from PR #640. The only potential **breaking change** is that I have changed the config property `ImageCacheSection.CachedRewritePathExpiry` (that I created in #640) to take a `TimeSpan` (and thus a `TimeSpan` literal like `0:1:0` for 1 minute) instead of an `int`. This uses a [`PositiveTimeSpanValidator`](https://msdn.microsoft.com/en-us/library/system.configuration.positivetimespanvalidatorattribute%28v=vs.110%29.aspx), so will presumably fail with a `ConfigurationErrorsException` if the config contains an integer literal.

**Note:** I have not been able to compile this to confirm, as the project has been upgraded to C# 7.0 with new-style [out variables](https://blogs.msdn.microsoft.com/dotnet/2017/03/09/new-features-in-c-7-0/), so will not build in Visual Studio 2015 and C# 6.0. I can confirm, however, that the only build errors are related to out variables, so the change should build correctly in Visual Studio 2017 and C# 7.0.

### DIFF
--- a/src/ImageProcessor.Web/Caching/CacheIndexer.cs
+++ b/src/ImageProcessor.Web/Caching/CacheIndexer.cs
@@ -56,21 +56,24 @@ namespace ImageProcessor.Web.Caching
         }
 
         /// <summary>
-        /// Adds a <see cref="cachedImage"/> to the cache.
+        /// Adds a <see cref="CachedImage"/> to the cache.
         /// </summary>
         /// <param name="cachedImage">
         /// The cached image to add.
         /// </param>
-        /// <param name="expiry">
-        /// The number of minutes to cache the image, defaults to 1.
+        /// <param name="expiration">
+        /// A <see cref="TimeSpan"/> defining the sliding expiration duration, defaults to zero
         /// </param>
         /// <returns>
         /// The value of the item to add or get.
         /// </returns>
-        public static CachedImage Add(CachedImage cachedImage, int expiry = 1)
+        public static CachedImage Add(CachedImage cachedImage, TimeSpan expiration = default(TimeSpan))
         {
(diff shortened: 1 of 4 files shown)


owenblacker:
@JimBobSquarePants Just to note, the AppVeyor CI script has run successfully, but it is failing to report here that it too cannot build C# 7.0:

> MSBUILD : error MSB1040: ToolsVersion is not valid. The tools version "15.0" is unrecognized. Available tools versions are "12.0", "14.0", "2.0", "3.5", "4.0".
Switch: 15.0
Error: 1/3/2018 3:09:51 PM: 
 \+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [<<==>>] Exception: Exec: Error executing command `Invoke-Expression $buildCommand`.
Collecting artifacts...
**No artifacts found matching 'build\_BuildOutput\NuGets\*.nupkg' path**

It might be worth editing the build script so that error `MSB1040` causes it to fail the build.

JimBobSquarePants:
Thanks @owenblacker for the PR and sorry for the delay!

I'm gonna update the build environment so it runs in VS2017 so that problem will go away.

PR #652, opened 2018-02-28
add default value for GetExtension method

perosb:
the inline documentation of GetExtension says default to "string.Empty", but the code returns "jpg".

In my opinion string.Empty actually sounds more feasible.

This commits add a parameter with default value "jpg" to make it easier to check if the requested input actually is supported or not by passing f.ex. null or string.Empty as default value.

### Prerequisites


JimBobSquarePants:
Thanks for this 👍 

I've pulled your PR but updated it to use an overload instead of a default value. This ensures that we don't break calling code through a method signature change. 

```
public string GetExtension(string fullPath, string queryString)
{
    return this.GetExtension(fullPath, queryString, "jpg");
}

public string GetExtension(string fullPath, string queryString, string defaultExtension)
{
    // Implementation
}
```

PR #654, opened 2018-03-08
Fix cache break preventing cache from being used again

### LINKED ISSUE #653: Updating a file successfully breaks cache, but doesn't resume using newly cached image
### Description
When updating an image on disk, the ImageProcessor succesfully returns the new image after processing and caching the new image.

However, all subsequent loads will miss the cache and process an entirely identical image every time.

I've made a fork, and I'm going to make a few changes that I've used in my dev environment, which I have verified as working. But the changes potentially have some far reaching implications so I want the owner of this project to be aware of what I'm doing and why. I'll make a PR related to this issue shortly

### Steps to Reproduce
1. Create and serve an image
    * Server runtime inspection shows the image being processed
2. Reload the image without browser cache to serve a new image from server cache
    * Server runtime inspection shows no processing, and the load is very fast
2. Modify the image in place
3. Reload the page without browser cache to serve the image anew
    * Server runtime inspection shows the image being processed
4. The image is successfully served as a new image, not the cached version
5. Reload the page once again
    * ISSUE: The server shows that the image is once again being processed, and this continues for every subsequent load indefinitely, this is incredibly slow, particularly in our case where we have hundreds of small sites split over very few servers

### System Configuration
- ImageProcessor version: Custom IImageCache implementation on ImageProcessor.Web - 4.9.1.2
- Environment (Operating system, version and so on): Dev server running on IIS 7 on Windows 10
- .NET Framework version: .NET 4.7

MellowArpeggiation:
Fixes #653

### Description
* Use the cached file modification time, rather than the creation time of the file
* Allow overwriting of existing cache files, whilst preserving the "hack" fix (catches errors rather than potentially throwing them)
* Simplify a DateTime.UtcNow reference

### DIFF
--- a/src/ImageProcessor.Web/Caching/DiskCache.cs
+++ b/src/ImageProcessor.Web/Caching/DiskCache.cs
@@ -180,13 +180,18 @@ namespace ImageProcessor.Web.Caching
 
             // This is a hack. I don't know why there would ever be double access to the file but there is
             // and it's causing errors to be thrown.
+            // Ensure that an attempt at overwrite is still made, so cached images can be updated
             // https://github.com/JimBobSquarePants/ImageProcessor/issues/629
-            if (!File.Exists(this.CachedPath))
+            try
             {
                 using (FileStream fileStream = File.Create(this.CachedPath, (int)stream.Length))
                 {
                     await stream.CopyToAsync(fileStream).ConfigureAwait(false);
                 }
+            }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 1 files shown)


MellowArpeggiation:
These modifications to the ImageProcessor.Web.Caching.DiskCache have been live for about 2 weeks for more than 350 of our sites, and it has measurably improved performance for us. Hopefully that works well enough as an integration test case!

JimBobSquarePants:
LGTM. Thanks for the fix! Much appreciated 👍

PR #660, opened 2018-04-06
Add MetaDataMode. Fix #585

### LINKED ISSUE #585: Add different options for preserveExifMetaData to preserve varying amount of data
### Description
There are scenarios where it is necessary to preserve different levels of Exif data, I would like to propose the following options are added to the preserveExifMetaData setting of processing.config

- all (also "true" for upgrade compatibility) - preserves all EXIF data
- copyrightOnly - preserves only the Copyright field
- minimal - preserves Copyright, DateTime , ExifUserComment, ImageDescription, and all fields relating to the camera, lens, exposure, etc
- minimalWithLocation - the same as minimal but including all GPS related fields

It could also be possible that a custom set of fields might be required, this would require a more complex configuration and may not be required frequently.

All fields can be viewed at https://github.com/JimBobSquarePants/ImageProcessor/blob/fef782988e5f39e536a1a303ed320dc061a68134/src/ImageProcessor/Imaging/MetaData/ExifPropertyTag.cs#L64

That's my opening sugesstion, hopefully others can provide a more conclusive list for each category

JimBobSquarePants:
### Prerequisites

### Description
Adds a new overload to `ImageFactory` that accept the `MetaDataMode` enum.

- None
- Copyright
- Geolocation
- CopyrightAndGeolocation
- All

Additionally the Web module `processing.config` now accepts a new optional property `metaDataMode` which when used in conjunction with the existing `preserveExifMetaData` property allows more fine grained control over what properties to preserve.

e.g
``` xml
<processing preserveExifMetaData="true" metaDataMode="Copyright" fixGamma="false" interceptAllRequests="true" allowCacheBuster="true">
```

All this is done in a fully backwards-compatible manner with both properties optional. 

@hartvig @Jeavon @drobar @PeteDuncanson @Matthew-Wise Please comment below with your thoughts. 

I think I've done enough to cover the basic requirements.
I don't want to add extreme customization as that allows the possibility for users to break things as `System.Drawing`s property items contain items other than EXIF metadata.

I'll do a PR to Umbraco to update version and config once this is merged and published.

### DIFF
--- a/src/ImageProcessor/Imaging/MetaData/ExifPropertyTagConstants.cs
+++ b/src/ImageProcessor/Imaging/MetaData/ExifPropertyTagConstants.cs
@@ -1,0 +1,87 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="ExifPropertyTagConstants.cs" company="James Jackson-South">
+//   Copyright (c) James Jackson-South.
+//   Licensed under the Apache License, Version 2.0.
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+
+namespace ImageProcessor.Imaging.MetaData
+{
+    /// <summary>
+    /// Contains constants grouping together common property items.
+    /// </summary>
+    public class ExifPropertyTagConstants
+    {
+        /// <summary>
+        /// Gets all required property items. The Gif format specifically requires these properties.
+        /// </summary>
+        public static readonly ExifPropertyTag[] RequiredPropertyItems = {
+            ExifPropertyTag.LoopCount, ExifPropertyTag.FrameDelay
+        };
+
+        /// <summary>
+        /// Gets all required property items plus geolocation specific property items. 
+        /// </summary>
+        public static readonly ExifPropertyTag[] GeolocationPropertyItems = RequiredPropertyItems.Union(new[]{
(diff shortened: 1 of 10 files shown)


src/ImageProcessor/Imaging/MetaData/ExifPropertyTagConstants.cs
```diff
@@ -0,0 +1,87 @@
...
+            ExifPropertyTag.GpsMapDatum,
+            ExifPropertyTag.GpsSpeed,
+            ExifPropertyTag.GpsSpeedRef,
+            ExifPropertyTag.GpsTrack,
+            ExifPropertyTag.GpsTrackRef,
+            ExifPropertyTag.GpsVer
+        }).ToArray();
+
+        /// <summary>
+        /// Gets all required property items plus copyright and geolocation specific property items.
```
Matthew-Wise: "and gelocation" needs removing from the comment here.

Other than that looks pretty solid to me. 👍

Jeavon:
Absolutely fabulous @JimBobSquarePants I think 90% of the time Copyright mode is all that is needed.

JimBobSquarePants:
Great! If no-one finds anything daft I'll ship it over the weekend.

JimBobSquarePants:
Turns out I can't do a PR against Umbraco cos they still target 4.5 🤷‍♂️

PR #666, opened 2018-04-18
fix Hue when not rotating

### LINKED ISSUE #665: Not possible to set Hue to 0 when not rotating
When you try to set the hue of an image to 0 while having rotate = false it is not changing the image because the function will always return.

See https://github.com/JimBobSquarePants/ImageProcessor/blob/2a747188b14cdac88272e1a7b97bf1df5fc91093/src/ImageProcessor/ImageFactory.cs#L925 and https://github.com/JimBobSquarePants/ImageProcessor/blob/2a747188b14cdac88272e1a7b97bf1df5fc91093/src/ImageProcessor/Processors/Hue.cs#L62

Adding an additional check if rotate = true will fix this.

MarcLandis:
If rotate is false it is not possible to set a hue of 0 as the function will return immediately.
This change will add a check if rotate is true and return only if so.

Fixes #665

### DIFF
--- a/src/ImageProcessor/ImageFactory.cs
+++ b/src/ImageProcessor/ImageFactory.cs
@@ -1013,7 +1013,7 @@ namespace ImageProcessor
         public ImageFactory Hue(int degrees, bool rotate = false)
         {
             // Sanitize the input.
-            if (degrees > 360 || degrees < 0 || degrees == 0)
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Ah that's great. Thanks, really appreciate it! 👍

Issue #669, opened 2018-05-22
Metadata is stripped after image is processed: ImageProcessor.Web 4.9.2.19


juliangrahame:
Hi James

Using: 

- ImageProcessor version: 2.6.1.19
- ImageProcessor.Web version: 4.9.2.19
- Environment: Windows 10, Windows Server 2012
- .NET Framework version: 4.6.1
- Umbraco 7.9.2

Got a call recently from a client (photographer) who stores licencing info in Image metadata according to the IPTC standard - http://www.controlledvocabulary.com/imagedatabases/iptc_naa.html. 
He mentioned that this metadata was being stripped out of any images that he uploaded to Umbraco. 

I tested this and indeed, before I upload the image to Umbraco, I can see the IPTC metadata section in PhotoShop - File Info (Windows properties never shows this stuff). After uploading and processing, when I save a copy off the site, the IPTC section is no longer there (all other metadata fields are persisted).

I have confirmed that this still happens after upgrade to latest version and setting `metaDataMode `to `All`.  

Perhaps you know more about IPTC than I do.  Personally I had never heard of IPTC and I am still pretty much in the dark. However,I am led to believe (by my client) that IPTC is very much a common standard so I have a few questions -

- Is this stripping of IPTC metadata by design?
- If it is then what was the reasoning? If not then is this something that is on the radar?
- Any plans to implement?

Can supply sample images if required.

Thanks in advance

Julian

juliangrahame:
Thanks @JimBobSquarePants 

Although I think there may be an issue with the version we have upgraded to (IP: 2.6.1.19, IP.Web version: 4.9.2.19) because after upgrading it turns out that no metadata at all seems to be persisted.  Apologies I hadn't realised this was happening.

I have this in my processing.config: 
`<processing preserveExifMetaData="true" metaDataMode="All" fixGamma="false" interceptAllRequests="false" allowCacheBuster="true">`

Was previously using the following versions which did persist a fair amount of metadata (Copyright camera information etc): 
ImageProcessor: 2.5.6
ImageProcessor.Web: 4.8.7

So with the latest version, we are not getting any metadata persisted at all after uploading to Umbraco and saving off the website.
J

JimBobSquarePants:
Can you do me a favour and check against the sample application in the repository? I want to make sure there's nothing overwriting setting in Umbraco.

juliangrahame:
Found it - I see I can set as parameter in call to ImageFactory.  

[...]

PR #670, opened 2018-05-28
#669 - metadata stripped after image processed 

juliangrahame:
ImageProessor.Web 4.9.2.19 whereby wrong MetaDataMode was being added in call to ImageFactory.

Reported in issue #669


JimBobSquarePants:
Looks absolutely perfect. Thanks for this! 💯

Issue #662, opened 2018-04-13
PostProcessor and access-issues


enkelmedia:
I'm using PostProcessor 1.3 on my Windows 2012 machine, and sometimes during startup I get this error in the log.

> RegisterExecutables 116 : Access to the path 'TruePNG.exe' is denied.,    at System.IO.Directory.DeleteHelper(String fullPath, String userPath, Boolean recursive, Boolean throwOnTopLevelDirectoryNotFound)
>   at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
>   at ImageProcessor.Web.Plugins.PostProcessor.PostProcessorBootstrapper.RegisterExecutables(). Inner: 

Followed by this
Unable to install postprocessor - No images will be post-processed. Unable to map working path for processors.

I'm not sure if its related to permissions since it works sometimes, and sometimes not. Reg. permissions - is there any guidance around if and how one might need to setup any custom permissions for PostProcessor to work?

UPDATE: 
I also get this exception every now and then
`RunProcess 151 : Unable to post process image for request https://www.domain.com/_files/objects/12792/my-png-image.png?width=300&height=225&mode=crop within 5000ms. Original image returned.`

And as far as I can tell it's only when a PNG-image i processed, I'm wondering if the timeout of 5 sec is enough if the PNG is a big foto for example? Or maybe the issue is related to the permissions stuff above.

I also get this error randomly
```
System.UnauthorizedAccessException: Access to the path 'C:\Windows\TEMP\tmp46F-out.png' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at ImageProcessor.Web.Plugins.PostProcessor.PostProcessor.PostProcessImage(HttpContext context, MemoryStream stream, String extension)
   at ImageProcessor.Web.Plugins.PostProcessor.PostProcessorApplicationEvents.PostProcess(Object sender, PostProcessingEventArgs e)
[...]

PR #674, opened 2018-06-13
Async PostProcessor without child processes

ronaldbarendse:
As mentioned in issue #662 I've rewritten the post processor to use async tasks and made sure it doesn't create child processes. All changes are within internal classes or private methods, so there should be no breaking changes.

I've also added a couple of performance improvements: it now uses async file/stream operations and doesn't write the temporary source file if no post processing tasks have to be performed (e.g. for WebP and BMP files).

I've added a package reference to https://github.com/jamesmanning/RunProcessAsTask, but if you'd like to remove this dependency, you could include it in ImageProcessor itself (it's only 2 classes and MIT licensed - so keep the copyright).

### DIFF
--- a/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
+++ b/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
@@ -116,9 +116,10 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
             {
                 ImageProcessorBootstrapper.Instance.Logger.Log(typeof(PostProcessor), $"Unable to post process image for request {context.Request.Unvalidated.Url} within {timeout}ms. Original image returned.");
             }
-            catch
+            catch (Exception ex)
             {
                 // Some security policies don't allow execution of programs in this way
+                ImageProcessorBootstrapper.Instance.Logger.Log(typeof(PostProcessor), ex.Message);
             }
             finally
             {
--- a/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
+++ b/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
@@ -15,6 +15,7 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
 	using System.Collections.Generic;
 	using System.Diagnostics;
 	using System.IO;
+	using System.Linq;
 	using System.Threading;
 	using System.Threading.Tasks;
 	using System.Web;
@@ -50,6 +51,9 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
 			int timeout = PostProcessorBootstrapper.Instance.Timout;
 			try
 			{
+				// Save source file length
+				long length = stream.Length;
+
 				// Create a temporary source file with the correct extension
 				string tempSourceFile = Path.GetTempFileName();
 				sourceFile = Path.ChangeExtension(tempSourceFile, extension);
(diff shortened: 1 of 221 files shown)


ronaldbarendse:
Here you go! 👍

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessorApplicationEvents.cs
```diff
@@ -55,7 +55,7 @@ public static void SetPostProcessingTimeout(int milliseconds)
         /// </param>
         private static void PostProcess(object sender, PostProcessingEventArgs e)
         {
-            e.ImageStream = PostProcessor.PostProcessImage(e.Context, e.ImageStream, e.ImageExtension);
+            e.ImageStream = PostProcessor.PostProcessImageAsync(e.Context, e.ImageStream, e.ImageExtension).Result;
```
ronaldbarendse: I just had to read the documentation and saw the `OnPostProcessing` event is awaitable (http://imageprocessor.org/imageprocessor-web/imageprocessingmodule/#events), so should we change this to an async method and await the `PostProcessImageAsync` call?

JimBobSquarePants: If we can yeah, that would be better.

ronaldbarendse: Just tested it and we can´t: mixing events and async just won't work together... In this case the `ImageStream` is already disposed, plus the event delegate is run in a fire-and-forget manner (so the image is already cached when the `PostProcessImageAsync` returns).

ronaldbarendse:
Just tested the `ImageProcessingModule.OnPostProcessing` event using async methods and that doesn't work, so the documentation should be updated: http://imageprocessor.org/imageprocessor-web/imageprocessingmodule/. Also, there´s no `CachedImagePath` property on the `PostProcessingEventArgs` class!

JimBobSquarePants:
It's been a while.... I'll need to do a lot of work to fix links in the website if I make a change now since github made a breaking change with jekyll.

JimBobSquarePants:
@ronaldbarendse Have you seen the latest changes to the develop branch. I've actually done away with child processes.

JimBobSquarePants:
I'll have another look then. Can you undo the Amazon changes? There's an open PR and you'll introduce conflicts.

https://github.com/JimBobSquarePants/ImageProcessor/pull/700

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -34,168 +37,135 @@ internal static class PostProcessor
         /// <param name="stream">The source image stream.</param>
         /// <param name="extension">The image extension.</param>
         /// <returns>
-        /// The <see cref="MemoryStream"/>.
+        /// The <see cref="MemoryStream" />.
         /// </returns>
-        public static MemoryStream PostProcessImage(HttpContext context, MemoryStream stream, string extension)
+        public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context, MemoryStream stream, string extension)
```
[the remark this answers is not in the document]
ronaldbarendse: There are use-cases for making the PostProcessor public, e.g. for optimizing uploaded images. In that case, we should probably ditch the `HttpContext` (or make it optional), because it's only used for logging...

JimBobSquarePants:
Merged! Thanks for pushing this!

PR #702, opened 2018-09-10
Added overlay color processor

ronaldbarendse:
Along the same lines as PR #628, I've added an `OverlayColor` processor, so you can overlay a (semi-transparent) color to the whole image.

I know you can just add an overlay in the HTML of a page, but because images with less colors are a _lot_ smaller (~50%), you'd be downloading unnessecary color data... We used this method (HTML overlay) on the 2nd and 3rd frontpage slide on https://www.tomatoworld.nl/en (so you have an idea of what the result will be).

I first tried to combine the background color and alpha processors, but that didn't work (the alpha processor only created a washed out image). Adding an overlay image (PNG filled with the semi-transparent color) with the exact same dimensions works, but adds a lot of extra work to add new dimensions and/or colors.

To get the solution to build, I merged the `Feature/AmazonS3Cache` branch, so you may want to do this yourself and cherry pick commit 7d4c364.

### DIFF
--- a/src/ImageProcessor.Web/Configuration/Resources/processing.config.transform
+++ b/src/ImageProcessor.Web/Configuration/Resources/processing.config.transform
@@ -46,6 +46,7 @@
         <setting key="VirtualPath" value="~/images/imageprocessor/overlay/"/>
       </settings>
     </plugin>
+    <plugin name="OverlayColor" type="ImageProcessor.Web.Processors.OverlayColor, ImageProcessor.Web"/>
     <plugin name="Pixelate" type="ImageProcessor.Web.Processors.Pixelate, ImageProcessor.Web"/>
     <plugin name="Quality" type="ImageProcessor.Web.Processors.Quality, ImageProcessor.Web" enabled="true"/>
     <plugin name="ReplaceColor" type="ImageProcessor.Web.Processors.ReplaceColor, ImageProcessor.Web"/>
--- a/src/ImageProcessor.Web/Processors/OverlayColor.cs
+++ b/src/ImageProcessor.Web/Processors/OverlayColor.cs
@@ -1,0 +1,73 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="OverlayColor.cs" company="James Jackson-South">
+//   Copyright (c) James Jackson-South.
+//   Licensed under the Apache License, Version 2.0.
+// </copyright>
+// <summary>
+//   Adds a color overlay to the current image.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace ImageProcessor.Web.Processors
+{
+	using System.Collections.Specialized;
+	using System.Drawing;
+	using System.Text.RegularExpressions;
+	using System.Web;
+
+	using ImageProcessor.Processors;
+	using ImageProcessor.Web.Helpers;
+
+	/// <summary>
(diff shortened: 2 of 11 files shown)


src/ImageProcessor.Web/Processors/OverlayColor.cs
```diff
@@ -0,0 +1,73 @@
...
+
+	/// <summary>
+	/// Adds a color overlay to the current image.
+	/// </summary>
+	public class OverlayColor : IWebGraphicsProcessor
+	{
+		/// <summary>
+		/// The regular expression to search strings for.
+		/// </summary>
+		private static readonly Regex QueryRegex = new Regex("color=[^&]", RegexOptions.Compiled);
```
ronaldbarendse: I'm not sure this Regex overlaps with the `bgcolor=` query, but while testing I haven't seen any errors indicating this would be an issue.

src/ImageProcessor/Processors/OverlayColor.cs
```diff
@@ -0,0 +1,78 @@
...
+		/// The processed image from the current instance of the <see cref="T:ImageProcessor.ImageFactory" /> class.
+		/// </returns>
+		public Image ProcessImage(ImageFactory factory)
+		{
+			Image image = factory.Image;
+
+			try
+			{
+				Color overlayColor = this.DynamicParameter;
+				if (overlayColor.A > 0)
```
ronaldbarendse: I added this check to short-circuit processing for fully transparent colors. While you wouldn't manually add this, it might be the case when generating the color overlays.

src/ImageProcessor.Web/Configuration/Resources/processing.config.transform
```diff
@@ -46,6 +46,7 @@
         <setting key="VirtualPath" value="~/images/imageprocessor/overlay/"/>
       </settings>
     </plugin>
+    <plugin name="OverlayColor" type="ImageProcessor.Web.Processors.OverlayColor, ImageProcessor.Web"/>
```
ronaldbarendse: Should this processor be enabled by default (like `BackgroundColor`)?

JimBobSquarePants:
I'm really not planning on adding any new features to the library? Perhaps this would be better as an extension?

ronaldbarendse:
I know you're focussing on `ImageSharp` and this library is in maintenance mode, but it's still in heavy use by e.g. Umbraco (also in the upcoming version 8) and ImageSharp is still in beta.

Having these additional features built-in would really be great and would avoid me creating a `ImageProcessor.Contrib` and `ImageProcessor.Web.Contrib` with all the version dependencies and maintenance for those additional packages 😛

If you need some help maintaining this library, just let me know!

JimBobSquarePants:
>Having these additional features built-in would really be great and would avoid me creating a ImageProcessor.Contrib and ImageProcessor.Web.Contrib with all the version dependencies and maintenance for those additional packages 😛

Nice try! It certainly would and when, eventually, you move on and I am left with extra code to maintain...

I think this is best as something separate. I won't be making any major changes to the main libraries since I'm essentially retiring them so you won't have updates to make.

JimBobSquarePants:
P.S Thanks for the offer to maintain

ronaldbarendse:
No problem! I'll add this processor to our internal set of packages and keep the branch, so other people can use the source if they'd like to 👍

PR #704, opened 2018-09-13
Fix for Shared Access Signature (SAS) Authorisation & new private blob support with Shared Access Signatures

idseefeld:
### Prerequisites

### Description
There are only two small changes in ImageProcessor.Web.Plugins.AzureBlobCache AzureBlobCache.cs to enable SAS token authorisation instead of shared access key authorisation. Especially for blob containers this allows a nice feature. You can create SAS tokens based on a previously defined _Access policy_. Which adds some security and makes changes of access rights possible. For more details please refer to the following doc:
[https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1](https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1)

**Microsoft Azure Storage Explorer** makes it easy to configure a certain blob container for this kind of authorisation.

Here is an example excerpt from ~/web.config (assuming web.config is used for configuration) which is the only part that changes for end users:

```
<add key="AzureBlobFileSystem.ConnectionString:media"
         value="BlobEndpoint=https://idseefeld.blob.core.windows.net;SharedAccessSignature=sv=2018-03-28&amp;si=media-rwdl&amp;sr=c&amp;sig=ztM9kj5Uz4fBIb2%2BJUBs%2BDwdNVX8%2Bjc%2Ballf%2B%2FktWLo%3D" /> 
```
I did manual testing, as I have no idea how do this with unit test :-(

This is done accordingly to UmbracoFileSystemProviders.Azure project.

As pull request is related to my solution for private blobs I have added my changes here.
Now the configuration options for Azure storage becomes quite complex and I'll start building a configuration and / or health check for this.

Happy to here your thougths. My config-files and code comments should explain how to use the new features. Have a look into the ImageProcessor.TestWebsite project for config files.

All new settings are optional and code is backward compatible.

Yours
Dirk

### DIFF
--- a/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
+++ b/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
@@ -60,6 +60,11 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
         private static CloudBlobContainer cloudSourceBlobContainer;
 
         /// <summary>
+        /// Cache item for Azure private blobs 'Shared Access Signature' (SAS)
+        /// </summary>
+        private static SASCache sasCache;
+
+        /// <summary>
         /// The cached root url for a content delivery network.
         /// </summary>
         private readonly string cachedCdnRoot;
@@ -106,7 +111,9 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
             if (cloudCachedBlobContainer == null)
             {
                 // Retrieve references to a container.
-                cloudCachedBlobContainer = CreateContainer(cloudCachedBlobClient, this.Settings["CachedBlobContainer"], BlobContainerPublicAccessType.Blob);
+                bool usePrivateContainer = this.Settings.ContainsKey("UsePrivateContainer") && !string.Equals(this.Settings["UsePrivateContainer"], "false", StringComparison.OrdinalIgnoreCase);
+                BlobContainerPublicAccessType accessType = usePrivateContainer ? BlobContainerPublicAccessType.Off : BlobContainerPublicAccessType.Blob;
+                cloudCachedBlobContainer = CreateContainer(cloudCachedBlobClient, this.Settings["CachedBlobContainer"], accessType);
             }
 
             if (cloudSourceBlobContainer == null)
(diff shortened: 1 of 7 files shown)


JimBobSquarePants:
@idseefeld I think this can be done without this. We simply pass the bytes to the response stream. Check out the latest code in the develop branch

idseefeld:
@JimBobSquarePants I have merged the latest develop branch an do no agree with the streaming approach. If I got it right streaming means: The webserver get's data on server side and delivers the same data to the client under websites domain. This makes the request to the CDN invisible for the client, but a lot of CDN benefits are lost.
On the other hand my approach wit attached sas token give you all CDN features like:
- data is directly transfered to the client with just the little redirect overhead
- as long we have to live with http1.1, the extra CDN domain gives us request performance
Yours
Dirk

idseefeld:
@JimBobSquarePants regarding a new AzureCloudImageService which you have mentioned somewhere, you are right. This service could use the same Azure APIs like the FileSystemProvider, because it needs the image data on server side to process them. 
So, to make it clear I was mainly talking about the the AzureBlobCache plugin project.
Which might have confused our conversation.
My changes in CloudImageService have just been a quick fix to deal with the cache issue.

idseefeld:
@JimBobSquarePants Just found your AzureImageService within AzureBlobCache plugin project (which makes sense;-) and fix an issue regarding container name configuration.
Now I hope this is ready for a release. Maybea beta like @Jeavon suggested for UmbracoFileSystemProviders.Azure project?

Jeavon:
@idseefeld did you manage to test with AzureImageService? I have been trying to test it but it's not happy with me. I get `ValueFactory attempted to access the Value property of this instance.` Config along these lines:
```
<service prefix="azure/" name="AzureImageService" type="ImageProcessor.Web.Plugins.AzureBlobCache.AzureImageService, ImageProcessor.Web.Plugins.AzureBlobCache">
            <settings>
                <setting key="StorageAccount" value="DefaultEndpointsProtocol=https;AccountName=;AccountKey="/>
                <setting key="Container" value=""/>
                <setting key="AccessType" value="Off"/>
            </settings>
        </service>
```

idseefeld:
@Jeavon I am not sure what you mean. I update the /bin of a new Umbraco v 7.12.3 website with my pull request DLLs and the latest version of https://github.com/JimBobSquarePants/UmbracoFileSystemProviders.Azure, configured everthing for private blobs (media & cache) - all manually. And it works like exspected. 
In this project all test passed without failures, but I cannot see nUnit test for the AzureImageService or AzureBlobCache plugins.
I also configured the ImageProcessor.TestWebsite and it seems to run without issues e.g. it generated http://localhost:61655/images/format-Penguins.png?crop=0,0,300,225 and redirects to my configured Azure cache blob: https://idseefeld.blob.core.windows.net/cache/e/1/1/6/1/6/e116167fae4269b23981c6bb6f2b3514ab1b8e2e.png?sv=2014-02-14&sr=c&sig=qeqwTbbpuj8xnEKvxKtSdHyWC6Ihz421vUEnR0P1qVg%3D&st=2018-10-27T10%3A09%3A47Z&se=2018-10-31T10%3A14%3A47Z&sp=r
When you call this url until 2018-10-31 morning, the SAS token should be valid. Removing the query parameter should prove the cache blob is private. Calling the url on November 1. should prove the SAS token has expired. 
In /config/imageprocessor/cache.config you may provide a new setting to set the expiration:
`<setting key="SASValidityInMinutes" value="5760"/> <!-- 4 days-->`

JimBobSquarePants:
I won't lie. I'm still struggling to understand why I need this? I can already serve private blobs.

idseefeld:
@JimBobSquarePants okay, there is the setting _StreamCachedImage_ in cache.config. Set to true image data is server by the original requested URL like www.mysite.com/media/1001/myimage.jpg?width=...
This means the the website host server requests the image data from Azure cache blob on server side using AzureImageService and then streams the data to the client.
The other way is to redirect to the cache blob URL which makes most out of CDN setups. But when cache blob is private e.g. to add some protection for closed communities (which is my real world requirement) you need this addiotional  SAS token querstring.
Here is an example:
Try 
```xml
https://ids-cdn.azureedge.net/cache/e/1/1/6/1/6/e116167fae4269b23981c6bb6f2b3514ab1b8e2e.png
```
you get an 404 response, which is a little misleading - actually it should be 403 AccessForbidden I think.
Now add 
```
?sv=2014-02-14&sr=c&sig=qeqwTbbpuj8xnEKvxKtSdHyWC6Ihz421vUEnR0P1qVg%3D&st=2018-10-27T10%3A09%3A47Z&se=2018-10-31T10%3A14%3A47Z&sp=r
```
Does it help or did I got something wrong?
@Jeavon what Do you say?

JimBobSquarePants:
>The other way is to redirect to the cache blob URL which makes most out of CDN setups. But when cache blob is private e.g. to add some protection for closed communities (which is my real world requirement) you need this addiotional SAS token querstring.

It doesn't though. 

If I could rewrite this from scratch I would avoid redirecting and always serve from a stream. 302's are super expensive and do not scale. If I could kill that 302 now, I would.

What users should be doing, and what this library should only support, is setting up a dedicated webapp as an endpoint to the CDN. 

That way you would request the image + command from the CDN, which then calls the dedicated app to process and cache the image. Should the CDN's own cache expire, we still have the cached image on the webapp server.

idseefeld:
I got your point. I have found @Jeavon 's AzureCDNToolkit, which seems to address this like you recommend - right. So this project would be the right place to handle the SAS tokens for private blobs.
I'll do some research...

gavinfaux:
`remote.axd` work's for this passing shared access key in query string e.g.:

`remote.axd/your.blob.core.windows.net/container-name/path/to/file.ext?sp=sharedaccesskey?width=600`

Currently this results in a redirect to cropped image on CDN 'https://your.azureedge.net/cache/x/y/z/random.png'.

Would be useful if cache could be bypassed just returning streamed image (for GDPR reasons; but I guess this can be done via other process).

JimBobSquarePants:
@gavinfaux I don't understand what you mean?

gavinfaux:
Ignore me, neither do I; trying to use ImageProcessor with private blob.

Issue #703, opened 2018-09-10
Add support for immutable cache header attribute


ronaldbarendse:
Adding an `immutable` attribute to the `Cache-Control` header ensures no requests have to be done to revalidate the cache when reloading a page, see: https://hacks.mozilla.org/2017/01/using-immutable-caching-to-speed-up-the-web/.

ImageProcessor already supports cache buster variables: https://github.com/JimBobSquarePants/ImageProcessor/blob/31a8c46a8556e3635273e025d435dbf593d093d3/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs#L700-L713

So should ImageProcessor add `immutable` to all responses when it detects a cache buster variable? Adding the attribute can be done using:
```c#
cache.AppendCacheExtension("immutable");
```

ronaldbarendse:
As a workaround, IIS outbound rewrite rules can be used:
```xml
<configuration>
  <system.webServer>
    <rewrite>
      <outboundRules>
        <rule name="AddCacheControlImmutable">
          <match serverVariable="RESPONSE_Cache_Control" pattern=".*" />
          <conditions logicalGrouping="MatchAny" trackAllCaptures="false">
            <add input="{QUERY_STRING}" pattern="(\?|&amp;)rnd=" />
            <add input="{QUERY_STRING}" pattern="(\?|&amp;)v=" />
          </conditions>
          <action type="Rewrite" value="public,max-age=31536000,immutable" />
        </rule>
      </outboundRules>
    </rewrite>
  </system.webServer>
</configuration>
```
This however overwrites the configured `browserMaxDays`, so it would be nice if it played well with the ImageProcessor configuration...

JimBobSquarePants:
I like this. Adds a feature without adding extra maintenance responsibility. 👍

ronaldbarendse:
I've improved the outbound IIS rewrite rule, so it only matches the exact `Cache-Control` header from ImageProcessor (seems to always be `public, must-revalidate, max-age=*`).

This also makes sure it only rewrites the header if the request is already publicly cachable and allows capturing the `max-age` value (so the configured `browserMaxDays` is honored and is kept in-sync with the `Expires` header). The query string conditions now also correctly match the cache buster value on the beginning of the string (because `{QUERY_STRING}` does not start with `?`).

```xml
<rule name="AddCacheControlImmutable">
[...]

PR #705, opened 2018-09-25
Adds immutable cache extension when cache busting variables are used

ronaldbarendse:
See issue #703.

### DIFF
--- a/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
+++ b/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
@@ -202,7 +202,15 @@ namespace ImageProcessor.Web.HttpModules
 
             cache.SetExpires(DateTime.Now.ToUniversalTime().AddDays(maxDays));
             cache.SetMaxAge(new TimeSpan(maxDays, 0, 0, 0));
-            cache.SetRevalidation(HttpCacheRevalidation.AllCaches);
+
+            if (ParseCacheBuster(context.Request.QueryString))
+            {
+                cache.AppendCacheExtension("immutable");
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
```diff
@@ -202,7 +202,15 @@ public static void SetHeaders(HttpContext context, string responseType, string[]
...
             cache.SetMaxAge(new TimeSpan(maxDays, 0, 0, 0));
-            cache.SetRevalidation(HttpCacheRevalidation.AllCaches);
+
+            if (ParseCacheBuster(context.Request.QueryString))
+            {
+                cache.AppendCacheExtension("immutable");
+            }
+            else
+            {
+                cache.SetRevalidation(HttpCacheRevalidation.AllCaches);
```
ronaldbarendse: Any reason why we should require stale resources to be re-validated?

ronaldbarendse: Looking into it, this seems to be the default behavior, but makes it unambiguous:
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#Cache_validation
- https://stackoverflow.com/a/8729854/96175

CloudVPS also recommends adding this to the `Cache-Control` header to make sure caches honor the `max-age` value: https://www.cloudvps.com/helpcenter/content-delivery-network-cdn/cdn-cache-control-and-invalidations.

So you'd still want to leave it in... I was just double checking 😃

JimBobSquarePants: Ace, let's merge your PR then 👍

PR #714, opened 2018-11-19
fix for container name path issue & additional request parameter for explicit redirect 

idseefeld:
### Prerequisites

- [x ] I have written a descriptive pull-request title
- [ x] I have verified that there are no overlapping [pull-requests](https://github.com/JimBobSquarePants/ImageProcessor/pulls) open
- [x ] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules :cop:.

### Description

Some configuration settings lead to a wrong request path. 
The additional explicit request query parameter in AzureBlobCache.cs should give the possibility to handle private blob protection in the frontend. For back office stream is sufficient(?). Actually private blobs make no sense when you stream them without further protection!

### DIFF
--- a/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
+++ b/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
@@ -324,7 +324,8 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
         /// </param>
         public override void RewritePath(HttpContext context)
         {
-            if (streamCachedImage)
+            bool requestRequiresRedirect = context.Request.Url.Query.ToLower().Contains("&redirect=true");
+            if (streamCachedImage && !requestRequiresRedirect)
             {
                 string blobPath = this.CachedPath.Substring(cloudCachedBlobContainer.Uri.ToString().Length + 1);
                 CloudBlockBlob blockBlob = cloudCachedBlobContainer.GetBlockBlobReference(blobPath);
--- a/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureImageService.cs
+++ b/src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureImageService.cs
@@ -15,6 +15,7 @@ namespace ImageProcessor.Web.Plugins.AzureBlobCache
     /// </summary>
     public class AzureImageService : IImageService
     {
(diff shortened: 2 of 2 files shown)


JimBobSquarePants:
Issues before PR's @idseefeld 

I don't even know reading this whether we're fixing an issue that there already exists functionality for.

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureBlobCache.cs
```diff
@@ -324,7 +324,8 @@ public override Task TrimCacheAsync()
         /// </param>
         public override void RewritePath(HttpContext context)
         {
-            if (streamCachedImage)
+            bool requestRequiresRedirect = context.Request.Url.Query.ToLower().Contains("&redirect=true");
```
JimBobSquarePants: We should never be reading the querystring here.

src/ImageProcessor.Web.Plugins.AzureBlobCache/AzureImageService.cs
```diff
@@ -59,7 +60,16 @@ public class AzureImageService : IImageService
         /// </returns>
         public async Task<byte[]> GetImage(object id)
         {
-            CloudBlockBlob blockBlob = this.blobContainer.GetBlockBlobReference(id.ToString());
+            CloudBlockBlob blockBlob = null;
```
JimBobSquarePants: I don't know what this is supposed to be fixing. I can always hit my container via the config.

JimBobSquarePants: Remember please that this was not built for Umbraco. If a change is required for Umbraco any fixes should be applied there.

idseefeld: Okay, I got that wrong. But are you sure this little extra check is not valuable for regular usage?
If you really do not want this in your package, I'll create a new PR for just the _AzureBlobCache_ request. Btw. are you fine with that one?

JimBobSquarePants: I don’t know what the cache change I supposed to achieve. We already have config for choosing whether to stream or redirect where both options are viable.

Issue #727, opened 2019-02-12
Query string corruption in ImageProcessingModule when using presets


ronaldbarendse:
### Description
The query string is corrupted when using multiple presets with the same prefix.

With these presets in `processing.config`:
```xml
<processing>
	<presets>
		<preset name="product" value="bgcolor=fff" />
		<preset name="product-sold" value="overlay=sold.png" />
	</presets>
	...
</processing>
```
The following query string is transformed/corrupted from `?anchor=center&mode=crop&width=200&height=200&preset=product&preset=product-sold` to `?anchor=center&mode=crop&width=200&height=200&bgcolor=fff&bgcolor=fff-sold`, but it should be transformed to: `?anchor=center&mode=crop&width=200&height=200&bgcolor=fff&overlay=sold.png`.

### Steps to Reproduce
Configure multiple presets with the same prefix and include multiple presets in the query string.

### System Configuration
This bug exists in the latest ImageProcessor.Web versions:
- ImageProcessor version: 2.7.0.100
- ImageProcessor.Web version: 4.10.0.100

See also: https://github.com/JimBobSquarePants/ImageProcessor/issues/710#issuecomment-461805908

PR #728, opened 2019-02-12
Query string corruption in ImageProcessingModule when using presets

ronaldbarendse:
This fixes issue #727.

### DIFF
--- a/src/ImageProcessor.Web/Helpers/UrlParser.cs
+++ b/src/ImageProcessor.Web/Helpers/UrlParser.cs
@@ -35,7 +35,7 @@ namespace ImageProcessor.Web.Helpers
             if (!string.IsNullOrWhiteSpace(servicePrefix))
             {
                 // Handle when prefix is contained in filename.
-                string[] split = Regex.Split(url, servicePrefix, RegexOptions.IgnoreCase);
+                string[] split = Regex.Split(url, Regex.Escape(servicePrefix), RegexOptions.IgnoreCase);
                 if (split.Length > 1)
                 {
                     url = string.Join(servicePrefix, split.Skip(1).ToArray()).TrimStart("?");
--- a/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
+++ b/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs
@@ -789,17 +789,13 @@ namespace ImageProcessor.Web.HttpModules
         {
             if (!string.IsNullOrWhiteSpace(queryString))
             {
-                foreach (Match match in PresetRegex.Matches(queryString))
+                queryString = PresetRegex.Replace(queryString, (match) =>
                 {
-                    if (match.Success)
-                    {
-                        string preset = match.Value.Split('=')[1];
+                    string preset = match.Value.Split('=')[1];
 
-                        // We use the processor config system to store the preset values.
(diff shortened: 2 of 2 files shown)


ronaldbarendse:
Also: what was the reason for using the `Regex.Replace` instead of a simple `String.Replace` in the previous code? `match.Value` could even contain valid regex patterns and cause unexpected replacements, so should be properly escaped first using [`Regex.Escape`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.escape).

This should probably still be done on this line: https://github.com/JimBobSquarePants/ImageProcessor/blob/21c4e284a80a8f3c63f89e0bdad2814740b4a0ba/src/ImageProcessor.Web/HttpModules/ImageProcessingModule.cs#L491

This could also be rewritten to use a combination of `String.IndexOf()` and `String.Substring()` calls, as the `String.Replace` method doesn't support case insensitivity...

JimBobSquarePants:
To answer your question. I wrote it about 4 years ago

ronaldbarendse:
I've looked into the query string processing code and it seems the casing of the URL/query string is kept intact, so ideally all matching/replacing should support different casings (especially because `Request.QueryString` is a `NameValueCollection` that already is case-insensitive and is used in combination with the regex matching)...

So all `Regex` instances in the `IWebGraphicsProcessor` classes should at least get a `RegexOptions.IgnoreCase` (and also `RegexOptions.CultureInvariant` and `RegexOptions.ExplicitCapture` while we're at it). Any regex pattern should also be escaped, like the `servicePrefix` in the `UrlParser.ParseUrl` method (as the dot in 'appdata.axd', 'database.axd' and 'remote.axd' currently means _any_ character):
https://github.com/JimBobSquarePants/ImageProcessor/blob/bab08e8f370989616d2283de1921cd41e3f04d41/src/ImageProcessor.Web/Helpers/UrlParser.cs#L38

I know this library is in maintenance mode, but it still widely used (e.g. all Umbraco sites). I don't mind putting some effort into this (basically adding some `RegexOptions` and checking for case sensitivity), if you're willing to accept more PRs for this 😉 I already spotted some performance improvements by moving Regex instances to static fields, etc.

JimBobSquarePants:
@ronaldbarendse As long as I don't need to do much work and the PRs give obvious benefit I'm happy to accept more.

ronaldbarendse:
I've added escaping to the `Regex.Replace()` (query string) and `Regex.Split()` (service prefix) patterns, so invalid input doesn't throw the following exception:
```
System.ArgumentException: parsing "center=0.52830188679245282,0.688&mode=crop&width=2560&height=1440&rnd=131842653730000000\" - Illegal \ at end of pattern.
 at System.Text.RegularExpressions.RegexParser.ScanBackslash()
 at System.Text.RegularExpressions.RegexParser.ScanRegex()
 at System.Text.RegularExpressions.RegexParser.Parse(String re, RegexOptions op)
 at System.Text.RegularExpressions.Regex..ctor(String pattern, RegexOptions options, TimeSpan matchTimeout, Boolean useCache)
 at System.Text.RegularExpressions.Regex.Replace(String input, String pattern, String replacement, RegexOptions options)
 at ImageProcessor.Web.HttpModules.ImageProcessingModule.<ProcessImageAsync>d__31.MoveNext()
```

This PR should just be a simple merge, as the changes are quite minor: the only real difference is that the service prefix (`IImageService.Prefix`) doesn't allow regex patterns anymore...

PR #737, opened 2019-04-02
Processors not matched when query string casing is different

ronaldbarendse:
In addition to PR #728, this changes the Regex options to ignore casing within the query string, use culture invariant and explicit capture to improve performance and correctly escapes input used in the patterns.

Because processors are matched using a Regex expression, but (if matched) use the case insensitive query string (`NameValueCollection`), some weird behavior can be experienced:
- `image.jpg?width=300` - resizes to 300
- `image.jpg?Width=300` - returns original image, no processing is done!
- `image.jpg?width=300&height=200` - resizes to 300x200
- `image.jpg?Width=300&height=200` - resizes to 300x200, because height is matched in the Regex and the `Resize` processor then parses the query string in a case insensitive way and can successfully get both width and height!

### DIFF
--- a/src/ImageProcessor.Web.Plugins.AmazonS3Cache/AmazonS3Cache.cs
+++ b/src/ImageProcessor.Web.Plugins.AmazonS3Cache/AmazonS3Cache.cs
@@ -40,7 +40,7 @@ namespace ImageProcessor.Web.Plugins.AmazonS3Cache
         /// <summary>
         /// The regular expression for parsing a remote uri.
         /// </summary>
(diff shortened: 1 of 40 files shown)


ronaldbarendse:
I've updated the branch, so it's now based on the latest develop branch. This fixes the described problem with different query string casings, includes additional Regex escaping and improved performance (by moving to static readonly fields and adding `CultureInvariant`, `ExplicitCapture` and `IgnoreCase` options as per best practice)!

Issue #713, opened 2018-11-06
PostProcessor fails to optimize pngs on Azure, although no errors occur


lars-erik:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageProcessor
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/JimBobSquarePants/ImageProcessor/issues) and [closed](https://github.com/JimBobSquarePants/ImageProcessor/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I can consistently reproduce pngs being 128 KB with a non-outputting process, but well optimized (99%) when the process has output. Very weird. :)

### Steps to Reproduce
Swap between these two branches:
https://github.com/lars-erik/ImageProcessor/tree/wtf-output
https://github.com/lars-erik/ImageProcessor/tree/develop

When the wtf-output dlls are used on the azure app, everything is fine.
When the develop dlls are used it does not err, but the pngs are 128 KB.

The change should not affect the process at all, given the WaitForExit, but it does!
The images are returned long before the 5 sec limit as well, so something is really weird.

Diff here:
https://github.com/lars-erik/ImageProcessor/commit/6b012d0ad0344f75b4dc9bd31fead23ca5443e22

The last File.AppendText calls didn't seem to put anything in D:\temp\local, but when remote debugging I could see the process output and everything was fine. Without it I have no idea what happens, and it doesn't work. Could be easier to figure out. :D

### System Configuration
Azure App Service

ImageProcessor, ImageProcessor.Web and ImageProcessor.Web.PostProcessor are built from this commit:
https://github.com/jimbobsquarepants/ImageProcessor/commit/eba1804d18ab492862c8bdf8da7c72a7e215bf85

- Environment (Operating system, version and so on):
Azure App Service

- .NET Framework version:
4.6.2

---

Have to stop this now. Have another major PR to attend to.

JimBobSquarePants:
Very odd..... I'll have to stick something online and run some tests.

lars-erik:
Give me a shout if I should dig up the originals our front-enders stuck in the storage account. Dunno if it's relevant that we use the our.umbraco.filesystemproviders.azure thingy. Freshest version.

lars-erik:
I wonder if it'll work with the events but no string concats. That shouldn't hit performance much, if at all.

[...]

PR #743, opened 2019-05-29
PostProcessor - Add option to disable TruePng via AppSetting

Jeavon:
This PR adds the option to disable TruePng execution by the PostProcessor plugin. This is required as some environments don't allow TruePng to execute properly. 

I and @lars-erik believe it's likely that TruePng is utilizing GDI32 which isn't available on platforms such as Azure WebApps, however, we are unable to prove this as the source code for TruePng isn't available. 

Whilst it's likely that there may be alternative tools that could replace TruePng this project is in soft archive and so I am not exploring this. 

As the PostProcssor doesn't already have a configuration file and I don't want to add one at this stage, in order to disable TruePng you will need to add an AppSetting:

`<add key="ImageProcessor.Web.PostProcessor.DisableTruePNG" value="true" />`

### DIFF
--- a/src/ImageProcessor.Web.Plugins.PostProcessor/ImageProcessor.Web.Plugins.PostProcessor.csproj
+++ b/src/ImageProcessor.Web.Plugins.PostProcessor/ImageProcessor.Web.Plugins.PostProcessor.csproj
@@ -34,6 +34,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="System" />
+    <Reference Include="System.configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.Web" />
     <Reference Include="System.Xml.Linq" />
--- a/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
+++ b/src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
@@ -19,8 +19,9 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
 	using System.Threading;
 	using System.Threading.Tasks;
 	using System.Web;
+    using System.Configuration;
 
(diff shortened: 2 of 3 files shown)


lars-erik:
@JimBobSquarePants I'm thinking I provided enough proof that TruePng fails on Azure when I was fiddling with it last time. Struggling to find the time to make another round just to provide "proof" (console output?).  
Reg. configuration, would it be better to add a static boolean one can set during startup instead?

As a reminder, here's PNGs ran through PostProcessor before and after the process events were added.
Before: https://www.markedspartner.no/media/1014/google-partner-liten.png  
After (1.4.0.100): https://markedspartner2017-staging.azurewebsites.net/media/1014/google-partner-liten.png  
I can reproduce the latter in the Kudu console by just running PngQuant and TruePng in order.

JimBobSquarePants:
With #713 I was able to replicate the issue locally and determined that the issue could be fixed by updating the version.

ronaldbarendse:
A good solution would be to open up the `PostProcessor` class, so it can be inherited and the `GetProcessStartInfos` method overridden with custom logic (e.g. calling base and removing `truepng.exe`). This would also allow post processing on non-processed images (outside ImageProcessor).

Disabling the default bootstrapping in `PostProcessorApplicationEvents` could be done using a config setting, so you can attach your own event on `ImageProcessingModule.OnPostProcessing` and also (temporarly) disable the post-processing.

Another option would be to make all `ProcessStartInfo` commands configurable, e.g. adding the following to the configuration:
```xml
<postprocessor>
   <extension name="png">
      <add fileName="pngquant.exe" arguments="--quality=90-99 --output &quot;{0}&quot; &quot;{1}&quot;" />
   </extension>
</postprocessor>
```

Issue #731, opened 2019-02-15
EntropyCrop error after update


srdjo:
I updated my ImageProcessor package via NuGet to latest version and now I get the followng error

![image](https://user-images.githubusercontent.com/4582491/52851905-85c80d80-3117-11e9-9833-9fe32d7ffef0.png)

?entropycrop=128

I am targeting NetFramework 4.7.1
ImageProcessor 2.7.0.100
ImageProcessor.Web 4.10.0.100

What might be the problem ?

JimBobSquarePants:
>What might be the problem ?

An invalid cast by the look of the stack trace.

However. From the Readme: 

>ImageProcessor is in Soft Archive Mode.
Focus for the ImageProcessor libraries has switched to a new library ImageSharp. As such, no new features will be added to this library. Bug fixes will continue to be added and only via PR.

srdjo:
Downgrading to ImageProcessor v2.6.2.25 fixesthe problem

JimBobSquarePants:
> Downgrading to ImageProcessor v2.6.2.25 fixes the problem

While introducing other issues. I never understand why people comment like this, not helpful to anyone.

stevenbarker:
Just created a PR now to fix this issue: #746

JimBobSquarePants:
Thanks, merged. Will do a release soon.

PR #746, opened 2019-06-25
Added byte cast to fix entropy crop web processor casting issue.

stevenbarker:
This is a fix for Issue #731 - I've added a cast to convert the literal 128 value into a byte within the ternary expression, this looks like it was originally removed in a bulk clean-up operation in commit [4af8de4a](https://github.com/JimBobSquarePants/ImageProcessor/commit/4af8de4a6376d9426fadd5c4596bc1de187899c0#diff-e424c650610a9bba9ed533296f341ca7).


JimBobSquarePants:
Ah brilliant thanks! :+1:

PR #750, opened 2019-06-28
Added Polly for retry/failover resilience on remote file handler to prevent 500 errors

ahwm:
### Prerequisites

- [X] I have provided test coverage for my change (where applicable)

### Description
This change adds failover resilience when resizing many (>10) remote images using `/remote.axd`. Currently, the library just fails and returns a 500 error and prevents the images from loading at all. This update fixes that issue and allows the user to configure it.

I made this update for a project we were working on that was loading up to 30+ (in some cases) images from a CDN that didn't have size options available. It's been running in production now with no issues.

### DIFF
--- a/src/ImageProcessor.Web/Configuration/Resources/security.config.transform
+++ b/src/ImageProcessor.Web/Configuration/Resources/security.config.transform
@@ -15,6 +15,8 @@
         <setting key="MaxBytes" value="4194304"/>
         <setting key="Timeout" value="3000"/>
         <setting key="Protocol" value="http"/>
+        <setting key="MaxRetryAttempts" value="3"/>
+        <setting key="PauseBetweenFailures" value="3"/>
       </settings>
       <whitelist>
       </whitelist>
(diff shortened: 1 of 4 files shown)


JimBobSquarePants:
@ahwm Sorry for the slow reply here, I've not been able to focus on the library as much as I'd like recently. I'll review as soon as possible. :+1:

ronaldbarendse:
I think we should make changes to `RemoteImageService` instead, so it's really easy to create a subclass with the Polly retry logic. That way, we're not adding a new dependency to the `ImageProcessor.Web` library and you can re-use the existing (maintained) code for fetching the remote files.

JimBobSquarePants:
>I think we should make changes to RemoteImageService instead, so it's really easy to create a subclass with the Polly retry logic. That way, we're not adding a new dependency to the ImageProcessor.Web library and you can re-use the existing (maintained) code for fetching the remote files.

@ronaldbarendse Agreed!

ahwm:
I'm all for that. I'm just not sure the best way to go about doing that. :)

ronaldbarendse:
@ahwm The best way is to create a `PollyRemoteImageService` class that inherits from `RemoteImageService` and try to make it work with minimal changes to `RemoteImageService` 😉 

Something like this should even work without making changes (not tested though!):
```c#
public class PollyRemoteImageService : RemoteImageService
{
    public override async Task<byte[]> GetImage(object id)
    {
        var retryPolicy = Policy.Handle<HttpRequestException>().WaitAndRetryAsync(3, i => TimeSpan.FromSeconds(3));

        return await retryPolicy.ExecuteAsync(async () =>
        {
            return await base.GetImage(id);
        }).ConfigureAwait(false);
    }
}
```

PR #753, opened 2019-07-10
It loads very slow on the first initialization

LavenderDaisy:
I use imageFactory.Load(imgPath); as below:
using (ImageFactory imageFactory = new ImageFactory(preserveExifData: true))
        {
            imageFactory.Load(imgPath);
            imageFactory.ReplaceColor(source, target, 128);

            using (MemoryStream ms = new MemoryStream())
            {
                imageFactory.Save(ms);
                return Image.FromStream(ms);
            }
        }

I have lots of assemblies in our system . It loads very slow on the first initialization,It takes 9965 ms to load all my assemblies which is not what I want . The init formats is enough for me .

ImageProcessor.Configuration.ImageProcessorBootstrapper.LoadSupportedImageFormats() 9833 ms
System.Linq.Enumerable.ToList(IEnumerable[TSource])   7724 ms
ImageProcessor.Configuration.ImageProcessorBootstrapper+<>c.<LoadSupportedImageFormats>b__18_0(Assembly)             7718ms
ImageProcessor.Common.Extensions.AssemblyExtensions.GetLoadableTypes(Assembly)  7718ms
System.Reflection.Assembly.GetTypes()                              7718ms

### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/JimBobSquarePants/ImageProcessor/pulls) open
- [ ] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules :cop:.

### DIFF
--- a/src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs
+++ b/src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs
@@ -76,7 +76,7 @@ namespace ImageProcessor.Configuration
         /// </summary>
         /// <param name="logger"></param>
         public void SetLogger(ILogger logger) => this.Logger = logger;
-
+        public static bool DefaultLoadAssembilies=true;
         /// <summary>
         /// Creates a list, using reflection, of supported image formats that ImageProcessor can run.
         /// </summary>
@@ -94,6 +94,8 @@ namespace ImageProcessor.Configuration
             Type type = typeof(ISupportedImageFormat);
             if (this.SupportedImageFormats == null)
             {
+                if(DefaultLoadAssembilies)
+                {
                 var availableTypes =
                     TypeFinder.GetAssembliesWithKnownExclusions()
                         .SelectMany(a => a.GetLoadableTypes())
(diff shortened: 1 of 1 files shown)


ronaldbarendse:
It's probably a good idea to make the bootstrapping/type finding pluggable, so this can be replaced with DI containers or other ways to load additional formats (e.g. using config files).

JimBobSquarePants:
@ronaldbarendse I’m actually dropping type loading entirely.

ronaldbarendse:
That might be the better option, as I see there's already a way to programmatically add new formats using `ImageProcessorBootstrapper.Instance.AddImageFormats()` and set a custom logger using `ImageProcessorBootstrapper.Instance.SetLogger()`.

This will be a breaking change though, as the `ImageProcessor.Plugins.WebP.WebPFormat` is loaded this way and other implementations rely on the auto-discovery (e.g. in Umbraco: https://github.com/umbraco/Umbraco-CMS/blob/v8/dev/src/Umbraco.Web/ImageProcessorLogger.cs).

JimBobSquarePants:
Which is fine. This would be for v3.

Note, I’m not currently planning on porting ImageProcessor.Web to the v3 API. Trying not to encourage System.Drawing on the web when better options are available

PR #754, opened 2019-07-10
It loads very slow on the first initialization 1

LavenderDaisy:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/JimBobSquarePants/ImageProcessor/pulls) open
- [ ] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules :cop:.

### DIFF
--- a/src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs
+++ b/src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs
@@ -77,6 +77,8 @@ namespace ImageProcessor.Configuration
         /// <param name="logger"></param>
         public void SetLogger(ILogger logger) => this.Logger = logger;
 
+        
+        public static bool DefaultLoadAssembilies = true;
         /// <summary>
         /// Creates a list, using reflection, of supported image formats that ImageProcessor can run.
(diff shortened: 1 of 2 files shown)


ronaldbarendse:
See my comment on https://github.com/JimBobSquarePants/ImageProcessor/pull/753#issuecomment-512131892, this should be made pluggable.

src/ImageProcessor/Configuration/ImageProcessorBootstrapper.cs
```diff
@@ -77,6 +77,8 @@ private ImageProcessorBootstrapper()
         /// <param name="logger"></param>
         public void SetLogger(ILogger logger) => this.Logger = logger;
 
+        
+        public static bool DefaultLoadAssembilies = true;
```
ronaldbarendse: You should change this to a property (as it's now a field).

ronaldbarendse: A better name would probably also be `LoadDefaultAssembilies`.

PR #756, opened 2019-07-16
Update retired WindowsAzure packages

Jeavon:
Updated retired Microsoft.WindowsAzure.Storage & Microsoft.WindowsAzre.ConfigurationManager to Microsoft.Azure.Storage.Blob & Microsoft.Azure.ConfigurationManager respectively

I have tested both the AzureBlobCache and AzureImageService


JimBobSquarePants:
Thanks @Jeavon I've been wanting to get around to this for a while!

Issue #778, opened 2019-12-20
Wrong image saved in cache


etharion:
### Versions
ImageProcessor: 2.8.0
ImageProcessor.Web: 4.11.0
ImageProcessor.Web.Config: 2.6.0
ImageProcessor.Web.PostProcessor: 1.5.0
Umbraco: 7.11.1
.NET Framework version: 4.6.1

### Description
We experience that the wrong images gets cached. For example in Umbraco you have the media archive with images uploaded on different nodes. When printing the HTML for the images, we first find the node and then the node contains information about which image file on the server to use. We use the img tag with a srcset, so we have multiple sizes (multiple cached images of the same image) for each image. It is here we experience one of the sizes will have another image from the media archive than the others, which is odd since the same node has been used to create the srcset. Haven't been able to find a stable way of reproducing this, but clear the cache and load images again (different images gets the wrong images from each cache clear). It is only happening to very few images, rough guess we experienced this on 5 out of 1000 images.

Examples of different sizes:
/media/6934/irish-turnout-black.jpg?width=1110&amp;rnd=201808100858&amp;heightratio=1&amp;quality=80&amp;bgcolor=ffffff

/media/6934/irish-turnout-black.jpg?width=750&amp;rnd=201808100858&amp;heightratio=1&amp;quality=80&amp;bgcolor=ffffff

A single of the sizes as shown here would show a different image apart from the others in the img tag and the querys would not be applied correctly either.
**/media/6934/irish-turnout-black.jpg?width=350&amp;rnd=201808100858&amp;heightratio=1&amp;quality=80&amp;bgcolor=ffffff**

The 'rnd'-query, we tried to get the last write date of the image file in hope that was enough to update the cache, but with no luck.

If needed I can send the project with a database. If anything else needed, I will try to answer. But this is mostly all we got about it.
### System Configuration
Windows 10 | Windows Server 2016
MSSQL 14

JimBobSquarePants:
This doesn't look  like an ImageProcessor issue to me. Perhaps talk to someone at Umbraco. 

![image](https://user-images.githubusercontent.com/385879/71297687-9a7cce00-23d8-11ea-99de-1e4b0ff1d38e.png)

Cached file paths are generated based on the full path and the querystring.

https://github.com/JimBobSquarePants/ImageProcessor/blob/61eb3195c37b840370a10126f57feb797741a306/src/ImageProcessor.Web/Caching/ImageCacheBase.cs#L155

[...]

PR #785, opened 2020-01-31
PostProcessor: generate better temporary file name and ensure all files are cleaned up

ronaldbarendse:
### Prerequisites

### Description
This PR fixes issue https://github.com/JimBobSquarePants/ImageProcessor/issues/778 by changing the way the temporary file name is generated, as `Path.GetTempFileName()` combined with changing the extension might cause the same name to be used concurrently and result in overwritten files.

It also reduces file copy actions and ensures all temporary files are always cleaned up.


src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -53,13 +53,11 @@ public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context
...
-                // Create a temporary source file with the correct extension
-                var tempSourceFile = Path.GetTempFileName();
-                sourceFile = Path.ChangeExtension(tempSourceFile, extension);
-                File.Move(tempSourceFile, sourceFile);
-
-                // Give our destination file a unique name
-                destinationFile = sourceFile.Replace(extension, "-out" + extension);
+                // Get temporary file names
+                var tempPath = Path.GetTempPath();
+                var tempFile = Path.GetFileNameWithoutExtension(Path.GetRandomFileName());
```
ronaldbarendse: Instead of using [`Path.GetTempFileName()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename) that creates predictable file names (and has a 65k file limit), this is using [`Path.GetRandomFileName()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getrandomfilename) to generate a cryptographically strong, random file name.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -80,6 +78,7 @@ public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context
                             if (File.Exists(destinationFile))
                             {
                                 File.Copy(destinationFile, sourceFile, true);
+                                File.Delete(destinationFile);
```
ronaldbarendse: This ensures the process always starts with only a source file, so the processor doesn't need to forcefully overwrite an existing destination file and if it doesn't generate a destination file (like pingo), ensures the processed file is used in the end (when multiple processors are executed).

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -99,16 +98,17 @@ public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context
                         }
                     }
 
-                    // Save result
+                    // Use source file if no destination file is created
                     if (!File.Exists(destinationFile))
                     {
-                        File.Copy(sourceFile, destinationFile, true);
+                        destinationFile = sourceFile;
```
ronaldbarendse: No need to copy the file, as we can just use the source file path as destination 😄

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -128,25 +128,22 @@ public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context
             finally
             {
                 // Always cleanup files
-                try
+                foreach (var fileName in new[] { sourceFile, destinationFile }.Where(f => f != null))
```
ronaldbarendse: The iterator might have some overhead, but this requires less duplicated code 😉

ronaldbarendse:
Disclaimer: this needs more testing to make sure it also keeps working in high concurrency environments, although it should at least work better than before :wink:

JimBobSquarePants:
Thanks @ronaldbarendse I'll review this ASAP.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: Or should we delete and move/rename, as that would probably cause less disk IOs?

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -99,16 +98,17 @@ public static async Task<MemoryStream> PostProcessImageAsync(HttpContext context
...
-                    // Save result
+                    // Use source file if no destination file is created
                     if (!File.Exists(destinationFile))
                     {
-                        File.Copy(sourceFile, destinationFile, true);
+                        destinationFile = sourceFile;
                     }
 
+                    // Save result
                     var result = new PostProcessingResultEventArgs(destinationFile, length);
```
ronaldbarendse: Any reason why we're using the additional `PostProcessingResultEventArgs` class, instead of getting the result file size and check whether it's smaller than the source (but not zero) right here?

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: I've decided better performance over duplicated code, so this is converted back (although it still deletes the destination file if an exception occurred while deleting the source file first).

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: I've removed the use of `PostProcessingResultEventArgs` and obsoleted it (as it was publicly exposed). We could also just remove the class completely, as it's probably not used by external code anyway.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: The latest changes use `FileInfo` for all IOs and keeps track whether it requires moving the file or not 👍

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -41,116 +41,142 @@ internal static class PostProcessor
...
+                                    {
+                                        // Use destination file as new source (for the next process)
+                                        destinationFileInfo.MoveTo(sourceFile);
+                                    }
+
+                                    // Swap source for destination
+                                    sourceFileInfo = destinationFileInfo;
+
+                                    // Try to keep the file in memory and ensure it's not indexed
+                                    sourceFileInfo.Attributes |= FileAttributes.Temporary | FileAttributes.NotContentIndexed;
```
ronaldbarendse: I'm not sure this makes any difference, as the file is already written. It probably depends whether the file is already flushed to disk, so it doesn't hurt to set it anyway.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -41,116 +41,142 @@ internal static class PostProcessor
...
+                        var sourceFileInfo = new FileInfo(sourceFile);
+                        using (var fileStream = sourceFileInfo.Create())
                         {
-                            // Use destination file as new source (if previous process created one).
-                            if (File.Exists(destinationFile))
-                            {
-                                File.Copy(destinationFile, sourceFile, true);
-                            }
+                            // Try to keep the file in memory and ensure it's not indexed
+                            sourceFileInfo.Attributes |= FileAttributes.Temporary | FileAttributes.NotContentIndexed;
```
ronaldbarendse: No need to flush the data back to mass storage and to index it, see: https://docs.microsoft.com/en-us/dotnet/api/system.io.fileattributes.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -41,116 +41,142 @@ internal static class PostProcessor
...
                     }
                 }
-                catch
-                {
-                    // Normally a no no, but logging would be excessive + temp files get cleaned up eventually.
-                }
             }
 
             // ALways return stream (even if it's not optimized)
-            stream.Position = 0;
```
ronaldbarendse: The stream is only altered within the try-catch-finally, so setting the position back to the beginning only has to be done in the finally-statement.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
```diff
@@ -41,116 +41,142 @@ internal static class PostProcessor
...
-                    {
-                        File.Copy(sourceFile, destinationFile, true);
-                    }
 
-                    var result = new PostProcessingResultEventArgs(destinationFile, length);
-                    if (result.ResultFileSize > 0 && result.Saving > 0)
-                    {
-                        using (var fileStream = File.OpenRead(destinationFile))
+                        // Refresh source file (because it's changed by external processes)
+                        sourceFileInfo.Refresh();
```
ronaldbarendse: This should not be required, as the `Exists` and `Length` properties are only accessed below, but this makes debugging a lot easier (because the debugger might already access these properties and return outdated information below).

ronaldbarendse: We can wrap this call within `#if DEBUG` and `#endif` statements to save an additional IO call 😄

etharion:
Is this something I should try run through on our solution too? We have a big load media.
Or is it ready for merge already?

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
JimBobSquarePants: Yeah let's do that. Not that this is the slow part of the process anyway but still, every little helps.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessingResultEventArgs.cs
```diff
@@ -22,6 +22,7 @@ namespace ImageProcessor.Web.Plugins.PostProcessor
 	/// <remarks>
 	/// Many thanks to Azure Image Optimizer <see href="https://github.com/ligershark/AzureJobs" />.
 	/// </remarks>
+	[Obsolete("This class was never publicly used by this library.")]
```
JimBobSquarePants: Let's just delete it. It definitely won't be used.

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessorApplicationEvents.cs
```diff
@@ -44,7 +44,7 @@ public static void SetPostProcessingTimeout(int milliseconds)
 		{
 			if (milliseconds > 0)
 			{
-				PostProcessorBootstrapper.Instance.Timout = milliseconds;
```
JimBobSquarePants: Oops!

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: It's now completely removed 👍

src/ImageProcessor.Web.Plugins.PostProcessor/PostProcessor.cs
ronaldbarendse: Looking at this again, I think it's best to keep the call, as you never know what the external process has done to the source file (it could've deleted it) and because it makes an important decision based on these properties (so it needs to be up-to-date).

JimBobSquarePants:
Will have a look at getting this merged and released tonight.

ronaldbarendse:
Because this contains a breaking change (by removing the public `PostProcessingResultEventArgs` class) and changes the way files are written, it should probably be released as version 2, right?

JimBobSquarePants:
> it should probably be released as version 2.0.0, right

Technically yes. I've played pretty loose and fast with versioning in the past due to upstream consumption limitations which was not helpful.

Issue #794, opened 2020-03-26
IO Errors with CacheTrimmer in Scaled Environment


jason-booth:
We're using ImageProcessor in an EPiServer site that is scaled across 3 servers who share one image cache. We've been seeing sporadic IO errors with CacheTrimmer like DirectoryNotFoundException and UnauthorizedAccessException (example below).

After talking with @JimBobSquarePants indeed the cache trimmer is not robust enough for a scaled environment.

StackTrace:    at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at ImageProcessor.Web.Caching.ImageCacheBase.CacheTrimmer.TimerRelease(Func`2 trimmer)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.TimerQueueTimer.CallCallback()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()

InnerException: System.IO.DirectoryNotFoundException

Message: Could not find a part of the path '\\<our shared drive>\image-cache\7\0\0\2\f'.

ronaldbarendse:
If multiple servers access the same cache directory, this could indeed happen (especially when you have a large amount of images, causing the trimming jobs to overlap). You can disable cache trimming entirely (`trimCache="false"` in the cache config) or only enable it on a single server as a temporary fix.

As trimming is done in a background task, the exception shouldn't affect the generation/caching of images: it could however stop trimming any further images.

The trimming code already does a check whether the directory exists, before enumerating the files:

https://github.com/JimBobSquarePants/ImageProcessor/blob/48bccd4e82e4ae51b53cdd79d9ecd9d34849e642/src/ImageProcessor.Web/Caching/DiskCache.cs#L224-L227

https://github.com/JimBobSquarePants/ImageProcessor/blob/48bccd4e82e4ae51b53cdd79d9ecd9d34849e642/src/ImageProcessor.Web/Caching/DiskCache.cs#L234-L236

[...]

PR #795, opened 2020-04-01
Fix race condition in DiskCache trimmer causing DirectoryNotFoundException

ronaldbarendse:
### Prerequisites

### Description
This fixes issue https://github.com/JimBobSquarePants/ImageProcessor/issues/794 by catching exceptions thrown while enumerating files from directories that are deleted while the cache trimming is executed (e.g. by another server running a concurrent cache trimming job or manually deleting the directory).

### DIFF
--- a/src/ImageProcessor.Web/Caching/DiskCache.cs
+++ b/src/ImageProcessor.Web/Caching/DiskCache.cs
@@ -211,64 +211,62 @@ namespace ImageProcessor.Web.Caching
 
             this.ScheduleCacheTrimmer(token =>
             {
-                string rootDirectory = Path.GetDirectoryName(this.CachedPath);
-
+                var rootDirectory = Path.GetDirectoryName(this.CachedPath);
                 if (rootDirectory != null)
                 {
-                    // Jump up to the parent branch to clean through the cache.
-                    // UNC folders can throw exceptions if the file doesn't exist.
-                    IEnumerable<string> directories = SafeEnumerateDirectories(validatedAbsoluteCachePath).Reverse();
-
+                    // Jump up to the parent branch to clean through the cache
+                    // UNC folders can throw exceptions if the file doesn't exist
+                    var directories = SafeEnumerateDirectories(validatedAbsoluteCachePath).Reverse().ToList();
                     foreach (string directory in directories)
                     {
-                        if (!Directory.Exists(directory))
-                        {
-                            continue;
-                        }
-
                         if (token.IsCancellationRequested)
                         {
                             break;
                         }
 
-                        IEnumerable<FileInfo> files = Directory.EnumerateFiles(directory)
-                                                               .Select(f => new FileInfo(f))
-                                                               .OrderBy(f => f.CreationTimeUtc);
-                        int count = files.Count();
-
-                        foreach (FileInfo fileInfo in files)
+                        try
                         {
-                            if (token.IsCancellationRequested)
+                            var files = Directory.EnumerateFiles(directory).Select(f => new FileInfo(f)).OrderBy(f => f.CreationTimeUtc).ToList();
+                            var count = files.Count;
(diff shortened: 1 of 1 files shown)


src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -211,64 +211,62 @@ public override Task TrimCacheAsync()
...
+                var rootDirectory = Path.GetDirectoryName(this.CachedPath);
                 if (rootDirectory != null)
                 {
-                    // Jump up to the parent branch to clean through the cache.
-                    // UNC folders can throw exceptions if the file doesn't exist.
-                    IEnumerable<string> directories = SafeEnumerateDirectories(validatedAbsoluteCachePath).Reverse();
-
+                    // Jump up to the parent branch to clean through the cache
+                    // UNC folders can throw exceptions if the file doesn't exist
+                    var directories = SafeEnumerateDirectories(validatedAbsoluteCachePath).Reverse().ToList();
```
ronaldbarendse, line 219: Even through `Reverse()` is using deferred execution, it still needs to enumerable all directories before it can return in reverse order. This makes sure it's executed on the same line and only does it once.

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -211,64 +211,62 @@ public override Task TrimCacheAsync()
...
                     foreach (string directory in directories)
                     {
-                        if (!Directory.Exists(directory))
-                        {
-                            continue;
-                        }
-
```
ronaldbarendse, line 228: I've removed this check, as the directory could still be deleted between now and enumerating the files below.

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -211,64 +211,62 @@ public override Task TrimCacheAsync()
...
+                        try
                         {
-                            if (token.IsCancellationRequested)
-                            {
-                                break;
-                            }
-
-                            try
+                            var files = Directory.EnumerateFiles(directory).Select(f => new FileInfo(f)).OrderBy(f => f.CreationTimeUtc).ToList();
+                            var count = files.Count;
```
ronaldbarendse, line 230: The previous `Count()` and foreach would enumerate the files twice, so converting it to a list first will ensure it's executed on the same line and only does it once.

src/ImageProcessor.Web/Caching/DiskCache.cs
```diff
@@ -211,64 +211,62 @@ public override Task TrimCacheAsync()
...
+                                {
+                                    // Log it but skip to the next file
+                                    ImageProcessorBootstrapper.Instance.Logger.Log<DiskCache>($"Unable to clean cached file: {fileInfo.FullName}, {ex.Message}");
+                                }
                             }
                         }
+                        catch (Exception ex)
+                        {
+                            // Log it but skip to the next directory
+                            ImageProcessorBootstrapper.Instance.Logger.Log<DiskCache>($"Unable to clean cached directory: {directory}, {ex.Message}");
```
ronaldbarendse, line 262: This is the same message as logged by `RecursivelyDeleteEmptyDirectories()`, but that won't cause any issues in debugging, as the exception message is also included.

ronaldbarendse:
Looks like `RecursivelyDeleteEmptyDirectories()` still throws `DirectoryNotFoundException`s, as the directory is deleted if it has no subdirectories/files and then tries to get its parent folder:

https://github.com/JimBobSquarePants/ImageProcessor/blob/8498d14566241a10c25df37b4c7fe79ad646b675/src/ImageProcessor.Web/Caching/DiskCache.cs#L577-L584

I think this whole method can be removed, as the cache trimmer already reverses all paths and therefore could do a single check (without recursively traversing to parent directories).

PR #796, opened 2020-04-06
Resize quality improvements

### LINKED ISSUE #792: Resize Quality can be improved.
See #790

Resize quality is far better in the v3 branch. We should backport some of that logic into the dev-v2 branch to improve quality.

e.g.
http://mb-website.com/prova.html

JimBobSquarePants:
### Prerequisites

### Description
Squeezed a bit of improved quality out of the Resize Processor. We're now producing the best quality that System.Drawing will allow. Fix #792 

In addition I replaced the inline dimension calculation code with the `ResizeHelper` used in the V3 branch and fixed a long standing issue parsing center coordinates in the web project. 

Light on tests but the code is heavily tested in the v3 branch and it's copy/paste so not going to bother adding as the infrastructure isn't present.


src/ImageProcessor/Imaging/Helpers/Adjustments.cs
```diff
@@ -329,7 +330,7 @@ private static byte[] GetLinearBytes()
             byte[] ramp = new byte[256];
             for (int x = 0; x < 256; ++x)
             {
-                ramp[x] = (255f * SRGBToLinear(x / 255f)).ToByte();
```
JimBobSquarePants, line 332: Improved the formula but this will never actually work all that well since the inverse LUT should be operating at a minimum of 12 bit precision, something we can never actually get without a custom resampler. That's why it's dropped in V3.

src/ImageProcessor/ImageFactory.cs
```diff
@@ -75,7 +75,7 @@ public class ImageFactory : IDisposable
         /// Whether to preserve exif metadata. Defaults to false.
         /// </param>
         public ImageFactory(bool preserveExifData = false)
-            : this(preserveExifData, true)
+            : this(preserveExifData, false)
```
ronaldbarendse, line 78: Why is the `fixGamma` parameter changed here?

src/ImageProcessor.Web/Helpers/QuerystringParser/Converters/PointFConverter.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// The <see cref="T:System.Globalization.CultureInfo"/> to use as the current culture. 
+        /// </param>
+        /// <param name="value">The <see cref="T:System.Object"/> to convert. </param>
+        /// <param name="propertyType">The property type that the converter will convert to.</param>
+        /// <exception cref="T:System.NotSupportedException">The conversion cannot be performed.</exception>
+        public override object ConvertFrom(CultureInfo culture, object value, Type propertyType)
+        {
+            object result = base.ConvertFrom(culture, value, propertyType);
+
+            return result is float[] list ? new PointF(list[0], list[1]) : result;
```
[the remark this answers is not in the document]
ronaldbarendse: The `PointConverter` class has the same problem BTW.

src/ImageProcessor/Imaging/ResizeHelper.cs
```diff
@@ -0,0 +1,394 @@
...
+                        targetY = (int)Math.Round((height - (sourceHeight * ratio)) / 2F);
+                        break;
+                }
+            }
+
+            // Target image width and height can be different to the rectangle width and height.
+            return (new Size(width, height), new Rectangle(targetX, targetY, targetWidth, targetHeight));
+        }
+
+        private static void ThrowInvalid(string message) => throw new InvalidOperationException(message);
```
ronaldbarendse, line 392: This seems like a silly helper method that will actually screw up the stack trace and line numbers.

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -102,67 +103,49 @@ public class ResizeLayer
         /// <summary>
         /// Gets or sets the center coordinates.
         /// </summary>
-        public float[] CenterCoordinates { get; set; }
+        public PointF? CenterCoordinates { get; set; }
```
ronaldbarendse, line 106: This is a binary breaking change (changing the type of a public property on a public class) and will require all code using it to be recompiled!

src/ImageProcessor/ImageFactory.cs
JimBobSquarePants: Because `LinearResize` is buggy an this triggers that over `CompositeResize`. An unfortunate change but unavoidable.

src/ImageProcessor/Imaging/ResizeHelper.cs
JimBobSquarePants: Actually static helper voids are very useful for code generation, especially on hotpaths where they allow inlining of calling methods. (this isn't one but every little helps.). By making them static you remove runtime overhead.

src/ImageProcessor/Imaging/ResizeLayer.cs
JimBobSquarePants: Hmmm... I doubt very much anyone would update the library without compilation.

ronaldbarendse: Not so fast! If another package uses ImageProcessor (e.g. Umbraco or your own utility package), that code won't automatically recompile and will require releasing a new version that only works on the latest ImageProcessor release.

This totally breaks backwards compatibility and will require the package maintainers to patch all supported versions. So this change can only be done when releasing ImageProcessor 3.

src/ImageProcessor/Imaging/ResizeHelper.cs
ronaldbarendse: But that's not the case here and it still messes with the stack trace: https://stackoverflow.com/a/55486648/96175.

src/ImageProcessor/Imaging/ResizeLayer.cs
ronaldbarendse: I've seen these kind of breaking changes multiple times and they're quite hard to notice, as tests are always recompiled and will just use the new signatures. They all reverted the change and released a patched version:
- https://github.com/zzzprojects/html-agility-pack/issues/375
- https://github.com/umbraco/Umbraco-CMS/issues/5822

src/ImageProcessor/Imaging/ResizeHelper.cs
JimBobSquarePants: That's not the same thing. You're comparing a throw helper in a different class to a private static member of the same class. 

The example given would require the `DebuggerStepThroughAttribute` to be present in order not to disrupt the stack trace.

src/ImageProcessor/Imaging/ResizeLayer.cs
JimBobSquarePants: Thing is, I really don't care. I have no obligation to any of those projects.

src/ImageProcessor/Imaging/ResizeHelper.cs
ronaldbarendse: Just to clarify, I meant the call stack on the exception that's thrown... That would still include the `ThrowInvalid` method/line number, even when it's marked with the `DebuggerStepThroughAttribute` (or even `DebuggerHiddenAttribute`).

So, if removing the exception handling from the `CalculateTargetLocationAndBounds` method would indeed cause it to be inlined and enhance performance for this hotpath, that would certainly be a good justification 😉

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -102,67 +103,49 @@ public class ResizeLayer
...
+        /// <param name="other">An object to compare with this object.</param>
         /// <returns>
-        /// A 32-bit signed integer that is the hash code for this instance.
-        /// </returns>
+        ///   <see langword="true" /> if the current object is equal to the <paramref name="other" /> parameter; otherwise, <see langword="false" />.</returns>
+        public bool Equals(ResizeLayer other)
+            => other != null
+            && this.Size == other.Size
+            && this.MaxSize == other.MaxSize
+            && this.RestrictedSizes.SequenceEqual(other.RestrictedSizes)
```
ronaldbarendse, line 128: This can cause an `ArgumentNullException` if `this.RestrictedSizes` or `other.RestrictedSizes` is `null`. Even through the constructor sets this to an empty list, the property is publicly settable and could therefore be set to `null` later.

This would be the simple fix:
```suggestion
            && (this.RestrictedSizes ?? new List<Size>()).SequenceEqual(other.RestrictedSizes ?? new List<Size>())
```

ronaldbarendse: Or slightly more complex:
```suggestion
            && (this.RestrictedSizes == null || other.RestrictedSizes == null ? this.RestrictedSizes == other.RestrictedSizes : this.RestrictedSizes.SequenceEqual(other.RestrictedSizes))
```
So if either is `null`, they should both be `null` to be equal, otherwise they're both not `null` and the sequence should be equal 👍

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -102,67 +103,49 @@ public class ResizeLayer
...
-                hashCode = (hashCode * 397) ^ (int)this.ResizeMode;
-                hashCode = (hashCode * 397) ^ (int)this.AnchorPosition;
-                hashCode = (hashCode * 397) ^ this.Upscale.GetHashCode();
-                hashCode = (hashCode * 397) ^ (this.CenterCoordinates?.GetHashCode() ?? 0);
-                return (hashCode * 397) ^ this.AnchorPoint.GetHashCode();
-            }
+            int hashCode = -1040263854;
+            hashCode = (hashCode * -1521134295) + this.Size.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.MaxSize.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.RestrictedSizes.GetHashCode();
```
ronaldbarendse, line 142: This has the same problem as the `Equals` method and should be updated to this (to my best knowledge of how `GetHashCode` works):
```suggestion
            hashCode = (hashCode * -1521134295) + (this.RestrictedSizes == null ? 0 : this.RestrictedSizes.GetHashCode());
```

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -102,67 +103,49 @@ public class ResizeLayer
...
-                hashCode = (hashCode * 397) ^ (this.RestrictedSizes?.GetHashCode() ?? 0);
-                hashCode = (hashCode * 397) ^ (int)this.ResizeMode;
-                hashCode = (hashCode * 397) ^ (int)this.AnchorPosition;
-                hashCode = (hashCode * 397) ^ this.Upscale.GetHashCode();
-                hashCode = (hashCode * 397) ^ (this.CenterCoordinates?.GetHashCode() ?? 0);
-                return (hashCode * 397) ^ this.AnchorPoint.GetHashCode();
-            }
+            int hashCode = -1040263854;
+            hashCode = (hashCode * -1521134295) + this.Size.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.MaxSize.GetHashCode();
```
ronaldbarendse, line 141:
```suggestion
            hashCode = (hashCode * -1521134295) + (this.MaxSize == null ? 0 : this.MaxSize.GetHashCode());
```

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -102,67 +103,49 @@ public class ResizeLayer
...
+            hashCode = (hashCode * -1521134295) + this.RestrictedSizes.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.ResizeMode.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.AnchorPosition.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.Upscale.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.CenterCoordinates.GetHashCode();
+            hashCode = (hashCode * -1521134295) + this.AnchorPoint.GetHashCode();
```
ronaldbarendse, line 147:
```suggestion
            hashCode = (hashCode * -1521134295) + (this.AnchorPoint == null ? 0 : this.AnchorPoint.GetHashCode());
```

src/ImageProcessor/Imaging/ResizeLayer.cs
ronaldbarendse: The same applies to your own packages, although I see you've already updated ImageProcessor.Web in this PR to require the latest version (but I'm not sure if other packages/plugins are also affected by this).

Besides the binary breaking change, won't this also swap the Y,X notation that's currently used? That would be another functional breaking change that can't be fixed with a recompile or even simple code changes. URLs including this notation might be used/stored somewhere (hardcoded in HTML or in the database) and this might even cause wrong images to be returned from the cache (and because of browser caching can't be simply fixed by clearing all cached images)!

ronaldbarendse: If the coordinates are indeed swapped, this would also be a problem for ImageProcessor 3...

JimBobSquarePants: Ah, good point on Y, X. I forgot about that.

src/ImageProcessor.Web/Helpers/QuerystringParser/Converters/PointFConverter.cs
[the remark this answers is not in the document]
ronaldbarendse: And `SizeConverter` 😉

ronaldbarendse:
I've added some fixes and ensured `ResizeLayer` is backwards compatible again (no binary breaking change and no swapping of X/Y coordinates), but the `ResizeHelper` still uses the validated and unambiguous `PointF?` (currently only as internal property).

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -62,107 +46,152 @@ public class ResizeLayer
...
             this.ResizeMode = resizeMode;
             this.AnchorPosition = anchorPosition;
-            this.CenterCoordinates = centerCoordinates ?? new float[] { };
+            if (centerCoordinates != null && centerCoordinates.Length == 2)
+            {
+                this.Center = new PointF(centerCoordinates[1], centerCoordinates[0]);
+            }
             this.MaxSize = maxSize;
-            this.RestrictedSizes = restrictedSizes ?? new List<Size>();
+            this.RestrictedSizes = restrictedSizes;
```
ronaldbarendse, line 54: No need to initialize this to an empty list, right? Implementations that expect a non-null value would already be susceptible to `NullReferenceException`s, as the property could be set to `null` after initialization.

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -62,107 +46,152 @@ public class ResizeLayer
...
+        }
+
+        /// <summary>
+        /// Gets the center coordinates as <see cref="Nullable{PointF}" />.
         /// </summary>
-        public float[] CenterCoordinates { get; set; }
+        /// <value>
+        /// The center coordinates as <see cref="Nullable{PointF}" />.
+        /// </value>
+        public PointF? Center { get; set; }
```
ronaldbarendse, line 139: This ensures backwards compatibility: keep the original `float[] CenterCoordinates` as obsolete property, while making it easier for other implementations to use the coordinates from the `PointF? Center` property 👍

src/ImageProcessor/Imaging/ResizeLayer.cs
JimBobSquarePants: It could be set to null by a third part but that would be silly. You could add a backing field to the property to easily fix that though.

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -10,44 +10,28 @@
...
-        /// </param>
+        /// <param name="size">The <see cref="T:System.Drawing.Size" /> containing the width and height to resize the image to.</param>
+        /// <param name="resizeMode">The resize mode to apply to the resized image.</param>
+        /// <param name="anchorPosition">The anchor position to apply to the resized image.</param>
+        /// <param name="upscale">Whether to allow up-scaling of images.</param>
+        /// <param name="centerCoordinates">The center coordinates (Y,X).</param>
+        /// <param name="maxSize">The maximum size to resize an image to. Used to restrict resizing based on calculated resizing.</param>
+        /// <param name="restrictedSizes">The range of sizes to restrict resizing an image to. Used to restrict resizing based on calculated resizing.</param>
+        /// <param name="anchorPoint">The anchor point.</param>
         public ResizeLayer(
```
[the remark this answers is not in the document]
JimBobSquarePants: I wish I could remove the constructor. It's such bad code.

src/ImageProcessor.Web/Processors/Resize.cs
```diff
@@ -75,16 +75,27 @@ public int MatchRegexIndex(string queryString)
...
+                    if (center.HasValue)
+                    {
+                        // Swap X/Y for backwards compatibility
+                        center = new PointF(center.Value.X, center.Value.Y);
+                    }
+                }
+                else
+                {
+                    center = null;
+                }
```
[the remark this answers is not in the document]
ronaldbarendse: Let's actually swap the values :sunglasses: 

```suggestion
                PointF? center;
                if (queryCollection["center"] != null)
                {
                    center = QueryParamParser.Instance.ParseValue<PointF>(queryCollection["center"]);
                    if (center.HasValue)
                    {
                        // Swap X/Y for backwards compatibility
                        center = new PointF(center.Value.Y, center.Value.X);
                    }
                }
                else
                {
                    center = null;
                }
```

src/ImageProcessor/Imaging/ResizeLayer.cs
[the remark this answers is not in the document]
ronaldbarendse: Version 3 could drop the `float[]` completely and clean up this constructor 👍

src/ImageProcessor/Imaging/ResizeLayer.cs
ronaldbarendse: As the swapped coordinates are only part of ImageProcessor.Web and support for that has dropped for version 3 (at least for now), this shouldn't cause a problem.

src/ImageProcessor/Imaging/ResizeLayer.cs
ronaldbarendse: The property is only used in `Resizer` and that already has a null-check: https://github.com/JimBobSquarePants/ImageProcessor/blob/4af8de4a6376d9426fadd5c4596bc1de187899c0/src/ImageProcessor/Imaging/Resizer.cs#L536

This is the kind of breaking change that won't cause major problems (although it should be documented in the release notes), but improves performance, as it doesn't need to construct the empty list 😉

ronaldbarendse:
@JimBobSquarePants This is ready to merge! If you can do a last check, this can be released right away (I've already made a draft release containing all changes and versions to bump/release).

PR #797, opened 2020-04-09
Fix and cleanup GetHashCode and Equals

ronaldbarendse:
### Prerequisites

### Description

This fixes some possible `NullReferenceException`s in the `Equals` implementations and uses `ValueTuple` to generate the hash codes, as that's a lot easier, cleaner and heavily optimized:
- https://stackoverflow.com/a/4630550
- https://intellitect.com/overidingobjectusingtuple/
- https://montemagno.com/optimizing-c-struct-equality-with-iequatable/

### DIFF
--- a/src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
+++ b/src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
@@ -10,18 +10,14 @@
 
 namespace ImageProcessor.Plugins.Cair.Imaging
 {
+    using System;
     using System.Drawing;
 
     /// <summary>
     /// Encapsulates the properties required to resize an image using content aware resizing.
     /// </summary>
-    public class ContentAwareResizeLayer
+    public class ContentAwareResizeLayer : IEquatable<ContentAwareResizeLayer>
     {
-        /// <summary>
-        /// The expected output type.
-        /// </summary>
-        private OutputType outputType = OutputType.Cair;
-
         /// <summary>
         /// Initializes a new instance of the <see cref="ContentAwareResizeLayer"/> class.
         /// </summary>
@@ -32,6 +28,11 @@ namespace ImageProcessor.Plugins.Cair.Imaging
         {
             this.Size = size;
         }
+
+        /// <summary>
+        /// Gets or sets the size.
+        /// </summary>
+        public Size Size { get; set; }
 
         /// <summary>
         /// Gets or sets the content aware resize convolution type (Default ContentAwareResizeConvolutionType.Prewitt).
@@ -46,23 +47,7 @@ namespace ImageProcessor.Plugins.Cair.Imaging
         /// <summary>
         /// Gets or sets the expected output type.
         /// </summary>
-        public OutputType OutputType
-        {
-            get
-            {
-                return this.outputType;
-            }
-
-            set
-            {
-                this.outputType = value;
-            }
-        }
-
-        /// <summary>
-        /// Gets or sets the size.
-        /// </summary>
-        public Size Size { get; set; }
+        public OutputType OutputType { get; set; } = OutputType.Cair;
 
         /// <summary>
         /// Gets or sets the the file path to a bitmap file that provides weight indicators specified using
@@ -94,21 +79,23 @@ namespace ImageProcessor.Plugins.Cair.Imaging
         /// <returns>
         ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
         /// </returns>
-        public override bool Equals(object obj)
-        {
-            if (!(obj is ContentAwareResizeLayer resizeLayer))
-            {
-                return false;
-            }
+        public override bool Equals(object obj) => obj is ContentAwareResizeLayer contentAwareResizeLayer && this.Equals(contentAwareResizeLayer);
 
-            return this.Size == resizeLayer.Size
-                && this.ConvolutionType == resizeLayer.ConvolutionType
-                && this.EnergyFunction == resizeLayer.EnergyFunction
-                && this.OutputType == resizeLayer.OutputType
-                && this.Parallelize == resizeLayer.Parallelize
-                && this.Timeout == resizeLayer.Timeout
-                && this.WeightPath == resizeLayer.WeightPath;
-        }
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// true if the current object is equal to the <paramref name="other" /> parameter; otherwise, false.
+        /// </returns>
+        public bool Equals(ContentAwareResizeLayer other) => other != null
+            && this.Size == other.Size
+            && this.ConvolutionType == other.ConvolutionType
+            && this.EnergyFunction == other.EnergyFunction
+            && this.OutputType == other.OutputType
+            && this.WeightPath == other.WeightPath
+            && this.Parallelize == other.Parallelize
+            && this.Timeout == other.Timeout;
 
         /// <summary>
         /// Returns a hash code for this instance.
@@ -116,6 +103,6 @@ namespace ImageProcessor.Plugins.Cair.Imaging
         /// <returns>
         /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table. 
         /// </returns>
-        public override int GetHashCode() => (this.ConvolutionType, this.EnergyFunction, this.Parallelize, this.OutputType, this.Timeout, this.Size, this.WeightPath).GetHashCode();
+        public override int GetHashCode() => (this.Size, this.ConvolutionType, this.EnergyFunction, this.OutputType, this.WeightPath, this.Parallelize, this.Timeout).GetHashCode();
     }
 }
--- a/src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
+++ b/src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
@@ -88,22 +88,15 @@ namespace ImageProcessor.Plugins.Cair.Imaging
         public int Timeout { get; set; } = 60000;
 
         /// <summary>
-        /// Returns a value that indicates whether the specified object is an 
-        /// <see cref="ContentAwareResizeLayer"/> object that is equivalent to 
-        /// this <see cref="ContentAwareResizeLayer"/> object.
+        /// Determines whether the specified <see cref="System.Object" />, is equal to this instance.
         /// </summary>
-        /// <param name="obj">
-        /// The object to test.
-        /// </param>
+        /// <param name="obj">The <see cref="System.Object" /> to compare with this instance.</param>
         /// <returns>
-        /// True if the given object  is an <see cref="ContentAwareResizeLayer"/> object that is equivalent to 
-        /// this <see cref="ContentAwareResizeLayer"/> object; otherwise, false.
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
         /// </returns>
         public override bool Equals(object obj)
         {
-            ContentAwareResizeLayer resizeLayer = obj as ContentAwareResizeLayer;
-
-            if (resizeLayer == null)
+            if (!(obj is ContentAwareResizeLayer resizeLayer))
             {
                 return false;
             }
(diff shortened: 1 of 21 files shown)


JimBobSquarePants:
There's a package now for hash code generation. 

https://www.nuget.org/packages/Microsoft.Bcl.HashCode

src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
```diff
@@ -88,22 +88,15 @@ public OutputType OutputType
...
-        /// <param name="obj">
-        /// The object to test.
-        /// </param>
+        /// <param name="obj">The <see cref="System.Object" /> to compare with this instance.</param>
         /// <returns>
-        /// True if the given object  is an <see cref="ContentAwareResizeLayer"/> object that is equivalent to 
-        /// this <see cref="ContentAwareResizeLayer"/> object; otherwise, false.
+        ///   <c>true</c> if the specified <see cref="System.Object" /> is equal to this instance; otherwise, <c>false</c>.
         /// </returns>
         public override bool Equals(object obj)
```
JimBobSquarePants, line 97: I just use `/// <inheritdoc/>` for this kind of stuff now. It works well as of recent VS versions.

src/ImageProcessor/Imaging/Colors/HSLAColor.cs
```diff
@@ -242,37 +242,30 @@ public override string ToString()
...
         /// <returns>
-        /// A 32-bit signed integer that is the hash code for this instance.
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table. 
         /// </returns>
-        public override int GetHashCode()
-        {
-            Color thisColor = this;
-            return thisColor.GetHashCode();
-        }
+        public override int GetHashCode() => ((Color)this).GetHashCode();
```
JimBobSquarePants, line 268: This is very expensive for hashcode generation

src/ImageProcessor/Imaging/Colors/YCbCrColor.cs
```diff
@@ -168,37 +168,30 @@ public override string ToString()
...
         /// <returns>
-        /// A 32-bit signed integer that is the hash code for this instance.
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table. 
         /// </returns>
-        public override int GetHashCode()
-        {
-            Color thisColor = this;
-            return thisColor.GetHashCode();
-        }
+        public override int GetHashCode() => ((Color)this).GetHashCode();
```
JimBobSquarePants, line 194: Again very expensive

src/ImageProcessor/Imaging/Formats/FormatBase.cs
```diff
@@ -136,23 +136,16 @@ public override bool Equals(object obj)
                 return false;
             }
 
-            return this.MimeType.Equals(format.MimeType) && this.IsIndexed.Equals(format.IsIndexed);
+            return (this.MimeType == null || format.MimeType == null ? this.MimeType == format.MimeType : this.MimeType.Equals(format.MimeType))
```
JimBobSquarePants, line 139: Just use `==`.

JimBobSquarePants:
ValueTuple relies on the default comparer for each property to compare equality. Is that enough?

src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
```diff
@@ -88,22 +88,15 @@ public OutputType OutputType
...
         {
-            ContentAwareResizeLayer resizeLayer = obj as ContentAwareResizeLayer;
-
-            if (resizeLayer == null)
+            if (!(obj is ContentAwareResizeLayer resizeLayer))
```
JimBobSquarePants, line 99: I prefer the pattern now.

``` c#
=> (obj is ContentAwareResizeLayer resizeLayer) && this.Equal(resizeLayer); 
```

src/ImageProcessor/Imaging/ResizeLayer.cs
```diff
@@ -110,59 +112,39 @@ public class ResizeLayer
...
+        /// </returns>
+        public bool Equals(ResizeLayer other)
+            => other != null
+            && this.Size == other.Size
+            && this.MaxSize == other.MaxSize
+            && (this.RestrictedSizes == null || other.RestrictedSizes == null ? this.RestrictedSizes == other.RestrictedSizes : this.RestrictedSizes.SequenceEqual(other.RestrictedSizes))
+            && this.ResizeMode == other.ResizeMode
+            && this.AnchorPosition == other.AnchorPosition
+            && this.Upscale == other.Upscale
+            && (this.CenterCoordinates == null || other.CenterCoordinates == null ? this.CenterCoordinates == other.CenterCoordinates : this.CenterCoordinates.SequenceEqual(other.CenterCoordinates))
```
JimBobSquarePants, line 139:
``` c#
this.RestrictedSizes?.SequenceEqual(other.RestrictedSizes)
this.CenterCoordinates?.SequenceEqual(other.CenterCoordinates)
```

ronaldbarendse:
I can confirm using `ValueTuple` for equality or hash code generation of collections doesn't work correctly indeed. Even the current `GetHashCode` implementations return different hash codes when the `Equals` method returns `true`, because it uses `SequenceEqual` to compare the collections (and not the default comparer), but the collections hash code:

```c#
var a = new List<int> { 1, 2, 3 };
var b = new List<int> { 1, 2, 3 };
a.Equals(b); // False
a.SequenceEqual(b); // True
a.GetHashCode() == b.GetHashCode(); // False, as different instances return different hash codes, even if the items are the same
```

**Equal items should return the same hash code (although the reverse isn't always true because of hash collisions), which isn't always the case in the current codebase.**

Microsoft has added a `HashCode` helper class in .NET Standard 2.1 (.NET Core 3) and released a [NuGet package for .NET Standard 2.0 (.NET Framework 4.6.1)](https://www.nuget.org/packages/Microsoft.Bcl.HashCode) to make this a lot easier, but upgrading the target framework from .NET Framework 4.5.2 to 4.6.1 for this isn't justified.

A quick fix would be to ensure equality is checked individually per property (with support for null values and collections) and just skip the collections in `GetHashCode` (using the `ValueTuple` construct).

JimBobSquarePants:
> Microsoft has added a HashCode helper class in .NET Standard 2.1 (.NET Core 3) and released a NuGet package for .NET Standard 2.0 (.NET Framework 4.6.1) to make this a lot easier, but upgrading the target framework from .NET Framework 4.5.2 to 4.6.1 for this isn't justified.

We copied the source for HashCode into the the SixLabors codebase for unsupported frameworks and use type forwarding to avoid conflict. Perhaps we can simply do the same here.

https://github.com/SixLabors/SharedInfrastructure/blob/master/src/SharedInfrastructure/HashCode.cs

src/ImageProcessor/Imaging/Filters/Photo/MatrixFilterBase.cs
```diff
@@ -29,41 +30,36 @@ public abstract class MatrixFilterBase : IMatrixFilter
...
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// true if the current object is equal to the <paramref name="other" /> parameter; otherwise, false.
+        /// </returns>
+        public bool Equals(IMatrixFilter other) => other != null
+            && this.GetType() == other.GetType()
+            && this.Matrix == other.Matrix;
```
ronaldbarendse, line 55: Comparing `ColorMatrix` is always done using object equality. To make this work, we would need to check all 25 matrix values.

src/ImageProcessor/Imaging/Filters/Photo/MatrixFilterBase.cs
```diff
@@ -29,41 +30,36 @@ public abstract class MatrixFilterBase : IMatrixFilter
...
         /// </returns>
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                int hashCode = this.GetType().Name.GetHashCode();
-                return (hashCode * 397) ^ this.Matrix.GetHashCode();
-            }
-        }
+        public override int GetHashCode() => (this.GetType(), this.Matrix).GetHashCode();
```
ronaldbarendse, line 63: The `ColorMatrix` works like a collection, so it returns different hash codes for objects that are initialized with the same matrix/items.

ronaldbarendse:
I've updated the test to handle the fixed `ColorMatrix` comparison, as the equality was previously ensured by reusing the same instance (from the `ColorMatrixes` helper class). I'm not sure what the performance difference will be, but the matrices can now also be initialized within the matrix filters themselves.

@JimBobSquarePants If you can review whether the unresolved comments are okay, I think this is ready to get merged!

src/ImageProcessor.Plugins.Cair/Imaging/ContentAwareResizeLayer.cs
ronaldbarendse: Changing the XML documentation to use `<inheritdoc />` might be something for another PR, so we don't need to block this one.

ronaldbarendse:
Note that merging this after PR #796 will result in a conflict on `ResizeLayer`: make sure this is resolved with the new `Equals` and `GetHashCode` methods and that both use the new `Center` property:
```c#
public bool Equals(ResizeLayer other) => other != null
    && this.Size == other.Size
    && this.MaxSize == other.MaxSize
    && (this.RestrictedSizes == null || other.RestrictedSizes == null ? this.RestrictedSizes == other.RestrictedSizes : this.RestrictedSizes.SequenceEqual(other.RestrictedSizes))
    && this.ResizeMode == other.ResizeMode
    && this.AnchorPosition == other.AnchorPosition
    && this.Upscale == other.Upscale
    && this.Center == other.Center
    && this.AnchorPoint == other.AnchorPoint;

public override int GetHashCode() => (this.Size, this.MaxSize, this.ResizeMode, this.AnchorPosition, this.Upscale, this.Center, this.AnchorPoint).GetHashCode();
```

src/ImageProcessor/Imaging/CropLayer.cs
```diff
@@ -69,47 +84,34 @@ public CropLayer(float left, float top, float right, float bottom, CropMode crop
...
-                return false;
-            }
+        public override bool Equals(object obj) => obj is CropLayer cropLayer && this.Equals(cropLayer);
 
-            // Define the tolerance for variation in their values 
-            return Math.Abs(this.Top - cropLayer.Top) <= Math.Abs(this.Top * .0001)
-                   && Math.Abs(this.Right - cropLayer.Right) <= Math.Abs(this.Right * .0001)
-                   && Math.Abs(this.Bottom - cropLayer.Bottom) <= Math.Abs(this.Bottom * .0001)
-                   && Math.Abs(this.Left - cropLayer.Left) <= Math.Abs(this.Left * .0001)
-                   && this.CropMode.Equals(cropLayer.CropMode);
```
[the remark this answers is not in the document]
JimBobSquarePants: We don't need this. Comparing two identical single precision floating point values will lead to the same result on the same CPU. 

Check this `PointF` example.

https://github.com/dotnet/runtime/blob/5cab5f03bfd336e3e1d158398252313d207dd6cb/src/libraries/System.Drawing.Primitives/src/System/Drawing/PointF.cs#L81

src/ImageProcessor/Imaging/Formats/FormatBase.cs
```diff
@@ -123,36 +123,31 @@ public virtual Image Save(string path, Image image, long bitDepth)
...
-        /// A 32-bit signed integer that is the hash code for this instance.
+        /// A hash code for this instance, suitable for use in hashing algorithms and data structures like a hash table. 
         /// </returns>
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                int hashCode = this.MimeType.GetHashCode();
-                hashCode = (hashCode * 397) ^ this.IsIndexed.GetHashCode();
-                return (hashCode * 397) ^ this.Quality;
```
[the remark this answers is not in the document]
JimBobSquarePants: Tbh we probably don't even need equality and hashcode overrides for most classes.

src/ImageProcessor/Imaging/Filters/Photo/MatrixFilterBase.cs
```diff
@@ -29,41 +30,88 @@ public abstract class MatrixFilterBase : IMatrixFilter
...
+        /// <summary>
+        /// Indicates whether the current object is equal to another object of the same type.
+        /// </summary>
+        /// <param name="other">An object to compare with this object.</param>
+        /// <returns>
+        /// true if the current object is equal to the <paramref name="other" /> parameter; otherwise, false.
+        /// </returns>
+        public bool Equals(IMatrixFilter other) => other != null
+            && this.GetType() == other.GetType()
+            && (this.Matrix == null || other.Matrix == null ? this.Matrix == other.Matrix : (
```
JimBobSquarePants, line 55: Wow.... Always thought `ColorMatrix` was a struct.

PR #799, opened 2020-04-21
Add support for .NET Standard 2.1 and other frameworks

optimus-code:
### Prerequisites

- [N/A] I have verified that I am following matches the existing coding patterns and practise as demonstrated in the repository. These follow strict Stylecop rules :cop:.
- [N/A] I have provided test coverage for my change (where applicable)

### Description
This adds support for .NET Standard 2.1, 2.0 and .NET Framework 4.7.2. It's up to you whether you want to add these changes, it has been helpful with my project.


ronaldbarendse:
ImageProcessor only supports .NET Framework 4.5.2 and higher (so including 4.8, as its backwards compatible) and won't add support for other platforms, as it relies on `System.Drawing` that's available on Windows (and that's not available or buggy on other platforms).

If you need support on these platforms, I would recommend you to use another library (like [ImageSharp](https://github.com/SixLabors/ImageSharp), as noted in the readme).

